PR #20, opened 2018-03-17
Add name of map

Fizo55:


0Lucifer0:
we should think about a way to get all language easily

PR #22, opened 2018-03-17
[ADD] MultiLanguage & Change OpenNosCore to NosCore

Fizo55:


0Lucifer0:
The multilanguage need to be different from opennos to support every language easily

PR #36, opened 2018-04-07
Add resource FR for the parser

Fizo55:


0Lucifer0:
will accepte this one but note that the translation system will be change soon

PR #56, opened 2018-04-24
[ADD] Help command

Fizo55:


src/NosCore.Handler/CommandPacketHandler.cs
```diff
@@ -29,13 +33,45 @@ public CommandPacketHandler(ClientSession session)
         #region Methods
         public void Speed(SpeedPacket speedPacket)
         {
-            if (speedPacket.Speed >= 60)
+            if (speedPacket.Speed > 0 || speedPacket.Speed < 60)
             {
```
Elendan: Useless & not working condition

0Lucifer0: thank you seems my post is displaying without specifing  the line :/ this  is the same thing i told 33min ago

0Lucifer0:
seems my comments are not correctly displayed so click on the link before each comments to see it should be linked

src/NosCore.Handler/CommandPacketHandler.cs
```diff
@@ -29,13 +33,45 @@ public CommandPacketHandler(ClientSession session)
...
+                }
+                Session.Character.Speed = speedPacket.Speed;
+                Session.SendPacket(Session.Character.GenerateCond());
+            }
+            else
+            {
+                Session.SendPacket(Session.Character.GenerateSay(speedPacket.Help(), 10));
+            }
+        }
+
```
Elendan: In the Help command, maybe adding an authority check would be a good idea (at least I don't see any), to only display available commands to the user, as an example, a GameMaster doesn't need to see Administrator commands and so on

src/NosCore.Handler/CommandPacketHandler.cs
Elendan: As well, something good for the user (imo) would be to set the speed at 59 even though he entered 999, instead of a return

src/NosCore.Handler/CommandPacketHandler.cs
0Lucifer0: This is already in the packet itself

0Lucifer0: Oh actually no I didn’t understood what you mean could be usefull yes

erwan-joly:
will need a rebase and the application of those reviews

PR #60, opened 2018-04-25
[UPDATE] Speed command

Fizo55:


src/NosCore.Handler/CommandPacketHandler.cs
```diff
@@ -31,11 +31,13 @@ public void Speed(SpeedPacket speedPacket)
         {
```
Elendan: That's a bit heavy, why not using ternary expression instead of duplicating code ? 

Session.Character.Speed = (speedPacket.Speed >= 60 ? 59 : speedPacket.Speed);

src/NosCore.Handler/CommandPacketHandler.cs
```diff
@@ -29,13 +29,9 @@ public CommandPacketHandler(ClientSession session)
         #region Methods
         public void Speed(SpeedPacket speedPacket)
         {
-            if (speedPacket.Speed >= 60)
-            {
-                return;
-            }
-            Session.Character.Speed = speedPacket.Speed;
+            Session.Character.Speed = (speedPacket.Speed >= Convert.ToByte(60) ? Convert.ToByte(59) : speedPacket.Speed);
```
erwan-joly: a cast to byte here would be better, shouldn't be nescessary you can check if Speed and packet.speed are bytes (they should be)

PR #75, opened 2018-05-08
Update commands

Victim-NH:
Insert AddMonster in commandpacketcontroller.cs

### DIFF
--- a/src/NosCore.Handler/CommandPacketController.cs
+++ b/src/NosCore.Handler/CommandPacketController.cs
@@ -17,43 +17,43 @@ namespace NosCore.Controllers
         }
         
     public void AddMonster(AddMonsterPacket addMonsterPacket)
-+        {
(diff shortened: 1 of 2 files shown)


Fizo55:
Where is the AddMonsterPacket ?

Victim-NH:
Entered now!

Fizo55:
"LogHelper.Instance.InsertCommandLog(Session.Character.CharacterId, addMonsterPacket, Session.IpAddress);" NosCore has not log for the commands ..

Fizo55:
GetNpc does not still exist on NosCore "ServerManager.Instance.GetNpc"

Fizo55:
To say that NosCore != OpenNos

Victim-NH:
Remove log and change GetNPC :)

0Lucifer0:
The goal is not to copy past things into noscore without understanding it.

PR #78, opened 2018-05-09
[ADD] enum for SayPacket color #74

Zagith:
added enum for SayPacker color and replaced all say color with this the enum

### DIFF
--- a/src/NosCore.Domain/SayColorType.cs
+++ b/src/NosCore.Domain/SayColorType.cs
@@ -2,16 +2,15 @@
 {
     public enum SayColorType
     {
-        DarkPurple = 0,
+        White = 0,
         LightGreen = 3,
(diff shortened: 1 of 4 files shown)


src/NosCore.Domain/SayColorType.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace NosCore.Domain
+{
+    public enum SayColorType
+    {
+        DarkPurple = 0,
+        LightGreen = 3,
+        DarkGrey = 4,
+        Red = 5,
```
0Lucifer0: 0 is white (only purple when GM)
5 is red whisper
8 is red
remove the 16

0Lucifer0:
fix #74

PR #125, opened 2018-07-01
[ADD] $ChangeRep Command

Fizo55:


0Lucifer0:
Moreover we need to avoid accessing someone else session in noscore.

PR #162, opened 2018-07-29
start adding inventory

0Lucifer0:


src/NosCore.Controllers/CommandPacketController.cs
```diff
@@ -53,6 +68,110 @@ public void Shout(ShoutPacket shoutPacket)
             ServerManager.Instance.BroadcastPackets(new List<PostedPacket>(new[] {sayPostedPacket, msgPostedPacket}));
         }
 
+        [UsedImplicitly]
+        public void CreateItem(CreateItemPacket createItemPacket)
```
Elendan: A lot of unnecessary nested conditions in this method

src/NosCore.Packets/ClientPackets/BIPacket.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class BIPacket : PacketDefinition
+    {
+        [PacketIndex(0)]
+        public PocketType PocketType { get; set; }
+
+        [PacketIndex(1)]
+        public byte Slot { get; set; }
+
+        [PacketIndex(2,IsOptional = true)]
+        public byte? Option { get; set; }
```
Elendan: Possible to make an enum for this ?

src/NosCore.Controllers/CommandPacketController.cs
0Lucifer0: Yes but admin command so not something really usefull

PR #180, opened 2018-08-09
Update login packet

Elendan:


src/NosCore.Packets/ClientPackets/NoS0575Packet.cs
```diff
@@ -19,6 +20,9 @@ public class NoS0575Packet : PacketDefinition
         public string Password { get; set; }
 
         [PacketIndex(3)]
+        public string LoginGuid { get; set; }
```
erwan-joly: use guid and make sure it works. Rename into something clearer ClientId for example

PR #190, opened 2018-08-20
Update LocalizedResources.es.resx

Adanlink:
146 - What is the utility of "<comment>check</comment>"?
370 - In that context "Drops" what mean? The items on the floor? The drops of the monsters?


0Lucifer0:
Its a comment someone added it has no goal. And yes item on the floor

PR #201, opened 2018-09-13
Add martial artist creation

Elendan:


src/NosCore.Controllers/CharacterScreenPacketController.cs
```diff
@@ -36,6 +36,96 @@ public CharacterScreenPacketController(WorldConfiguration worldConfiguration)
...
         {
         }
+
+        /// <summary>
+        ///     Char_NEW character creation character
+        /// </summary>
+        /// <param name="brawlerCreatePacket"></param>
+        public void CreateBrawlerCharacter(BrawlerCreatePacket brawlerCreatePacket)
+        {
+            if (Session.HasCurrentMapInstance)
```
0Lucifer0: there is something to check this on packets

PR #239, opened 2018-10-14
[FIX] Login authority

Fizo55:
- [x] authority < 0 can login
- [x] AuthorityType '3' isn't working (switch for '2' in DB)
- [x] Authority '1' Give [Support] prefix. [Support] rank is invisible to other players. It's also named in the character name (own player only).
- [x] support auth can't Shout.
- [x]  Authority '-1; -2' give Admin rank (with shout).
- [x] Authority '-3' mark as a normal player (but Shout works). Banned and closed accounts can be connected as usual.


0Lucifer0:
fix #237

src/NosCore.GameObject/ComponentEntities/Extensions/VisualEntityExtension.cs
```diff
@@ -66,7 +67,7 @@ public static InPacket GenerateIn(this ICharacterEntity visualEntity)
             return new InPacket
             {
                 VisualType = visualEntity.VisualType,
-                Name = visualEntity.Name,
+                Name = visualEntity.Authority == AuthorityType.Moderator ? "[Support]" + visualEntity.Name : visualEntity.Name,
```
Elendan: Don't hardcode the "Support", there is a language key for that

src/NosCore.GameObject/ComponentEntities/Extensions/VisualEntityExtension.cs
```diff
@@ -66,7 +68,7 @@ public static InPacket GenerateIn(this ICharacterEntity visualEntity)
             return new InPacket
             {
                 VisualType = visualEntity.VisualType,
-                Name = visualEntity.Name,
+                Name = visualEntity.Authority == AuthorityType.Moderator ? $"{visualEntity.Session.GetMessageFromKey(LanguageKey.SUPPORT)}" + visualEntity.Name : visualEntity.Name,
```
0Lucifer0: You don’t need the string part

PR #262, opened 2018-10-25
Add serilog

0Lucifer0:


erwan-joly:
https://nblumhardt.com/2016/03/reading-logger-configuration-from-appsettings-json/

PR #264, opened 2018-10-25
Martial artist creation

Elendan:


0Lucifer0:
will need to rebase and apply log change

PR #288, opened 2018-11-07
Added german translation

VIONYT:


src/NosCore.Shared/Resource/LocalizedResources.de.resx
```diff
@@ -1,120 +1,609 @@
...
+  <data name="USER_TELEPORTED_TO_NOT_EXISTING_MAP1" xml:space="preserve">
+    <value>Benutzer zu nicht existierender Karte teleportiert!</value>
+  </data>
+  <data name="VISUALTYPE_UNKNOWN1" xml:space="preserve">
+    <value>Unbekannter VisualType: {0}</value>
+  </data>
+  <data name="WRESTLER1" xml:space="preserve">
+    <value>Kämpfer</value>
+  </data>
+  <data name="WRONG_SELECTED_MAPID1" xml:space="preserve">
```
0Lucifer0: Why do you have 1 after each keys ?

VIONYT: Actually i don't know

0Lucifer0: Remove them or it won’t work

PR #334, opened 2018-12-02
French Localization update 2

Digitalis87:
Thanks Adanlink for stings context.

### DIFF
--- a/src/NosCore.Shared/Resource/LocalizedResources.fr.resx
+++ b/src/NosCore.Shared/Resource/LocalizedResources.fr.resx
@@ -453,7 +453,7 @@
   <data name="SURE_TO_DELETE" xml:space="preserve">
     <value>Êtes-vous sûr de vouloir supprimer ?</value>
   </data>
-  <data name="ACCEPTED_SHARE" xml:space="preserve">
+  <data name="Nope :D" xml:space="preserve">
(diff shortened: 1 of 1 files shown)


src/NosCore.Shared/Resource/LocalizedResources.fr.resx
```diff
@@ -453,7 +453,7 @@
   <data name="SURE_TO_DELETE" xml:space="preserve">
     <value>Êtes-vous sûr de vouloir supprimer ?</value>
   </data>
-  <data name="ACCEPTED_SHARE" xml:space="preserve">
+  <data name="Nope :D" xml:space="preserve">
```
erwan-joly: nope

src/NosCore.Shared/Resource/LocalizedResources.fr.resx
```diff
@@ -580,11 +580,9 @@
     <value>Ton niveau héroïque a augementé.</value>
   </data>
   <data name="MASTER_SERVER_PING" xml:space="preserve">
-    <value />
-    <comment>Need context for translate</comment>
+    <value>Ping réussi, le serveur master est actif.</value>
```
erwan-joly: actually it's sent on every ping even if not successful

PR #355, opened 2018-12-09
Add an object containing api routes

Elendan:


0Lucifer0:
i think this object belong to Shared also codacy comments

PR #379, opened 2018-12-23
Disconnect command

Fizo55:


0Lucifer0:
LGTM will wait for @Elendan review

src/NosCore.Controllers/CommandPacketController.cs
```diff
@@ -67,6 +67,32 @@ public CommandPacketController()
...
+        public void Kick(KickPacket kickPacket)
+        {
+            var servers = WebApiAccess.Instance.Get<List<ChannelInfo>>("api/channel");
+            ConnectedAccount account = null;
+            ServerConfiguration config = null;
+
+            foreach (var server in servers)
+            {
+                config = server.WebApi;
+                account = WebApiAccess.Instance.Get<List<ConnectedAccount>>("api/connectedAccount", config)
```
Elendan: You need to get out of the loop when the account is not null

src/NosCore.Controllers/CommandPacketController.cs
```diff
@@ -67,6 +67,36 @@ public CommandPacketController()
         {
         }
 
+        [UsedImplicitly]
+        public void Kick(KickPacket kickPacket)
+        {
+            var servers = WebApiAccess.Instance.Get<List<ChannelInfo>>("api/channel");
```
Elendan: Specify channel type (use .where)

PR #423, opened 2019-01-04
Add changeclass command

Fizo55:


src/NosCore.WorldServer/Controllers/StatController.cs
```diff
@@ -79,6 +80,13 @@ public IActionResult UpdateStats([FromBody] StatData data)
                     }
                     session.SetGold(data.Data);
                     break;
+                case UpdateStatActionType.UpdateClass:
+                    if ((byte) session.Class == data.Data)
```
Elendan: Maybe this check should be done in the ChangeClass method, this would seem more logical to me

src/NosCore.Packets/CommandPackets/ChangeClassPacket.cs
```diff
@@ -0,0 +1,41 @@
...
+using NosCore.Core.Serializing;
+using NosCore.Shared.Enumerations.Account;
+
+namespace NosCore.Packets.CommandPackets
+{
+    [PacketHeader("$ChangeClass", Authority = AuthorityType.GameMaster)]
+    public class ChangeClassPacket : PacketDefinition, ICommandPacket
+    {
+        [PacketIndex(0)]
+        [Range(0, 4)]
```
0Lucifer0: no need a range

src/NosCore.Packets/CommandPackets/ChangeClassPacket.cs
```diff
@@ -0,0 +1,41 @@
...
+        public byte ClassType { get; set; }
```
0Lucifer0: should be the class enum

src/NosCore.Packets/CommandPackets/ChangeClassPacket.cs
```diff
@@ -0,0 +1,41 @@
...
+
+        [PacketIndex(1, SerializeToEnd = true)]
+        public string Name { get; set; }
```
0Lucifer0: Optional

PR #429, opened 2019-01-07
WIP: Exchange tests

Elendan:


erwan-joly:
avoid fixing build and doing work in same PR ;) will merge this. Still need a lot more tests (controller need to be tested too in order to check that we are returning the correct message in every cases.) check Inventory tests for more details

PR #450, opened 2019-01-12
Dir packet implementation and removed useless "mv" packet

Adanlink:
0Lucifer0 helped me.


src/NosCore.Packets/ServerPackets/DirPacket.cs
```diff
@@ -0,0 +1,43 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+using NosCore.Core.Serializing;
+using NosCore.Shared.Enumerations;
+using System.ComponentModel.DataAnnotations;
+
+namespace NosCore.Packets.ClientPackets
+{
+    [PacketHeader("dir")]
+    public class DirPacket : PacketDefinition
```
Elendan: Why is there a second dir packet in the Client packets?

Elendan:
Why has the code been completely re indented ? can't review that

Adanlink:
Srry I needed to upload the files via webpage. And yeah, I messed up with the two packets. Dir_Packet.cs needs to go to the Server Packets and DirPacket.cs to the client one. When I arrive home I will fix it.

0Lucifer0:
This is unreviewable

PR #452, opened 2019-01-15
Added some translation

VIONYT:


0Lucifer0:
You definitely don’t want to push all those changes

PR #453, opened 2019-01-15
Added some translation

VIONYT:


0Lucifer0:
Parameter test is failing. You have some {0} non required

PR #456, opened 2019-01-15
Update Italian LocalizedResources

Wesker91:


0Lucifer0:
Bad parameters + JSon files

PR #461, opened 2019-01-17
Add Size variable to "c_mode" and the character, also add "$Size" command

Adanlink:


Elendan:
Size should not be in the c_mode packet, but un the char_sc packet, which is formatted this way:

char_sc [VisualType] [VisualId] [Size]

0Lucifer0:
Seems official has it in cmode 1 now but I was surprise

Adanlink:
I have a dilemma. With "char_sc" you are able to modify the size of any IAliveEntity (NPCs, Monsters, Players), but the problem is that the "in" packet only integrates the size when used in Players. So, I was thinking to add the size value to NPCs and Monsters too. The only problem is that when size of monsters isn't equal to 10 we would need to send the "char_sc" packet when the players change maps. I know it isn't used in official, but it lets you get some of the juice from NosTale that was never used. (So I would need to upgrade the $Size command to be able to change NPCs, Monsters, etc.)

0Lucifer0:
Im ok with this. Add size to alive entity and a check with the in if size != 10 send char_sc it will make gm able to call big monsters as boss for events

src/NosCore.GameObject/ComponentEntities/Interfaces/IVisualEntity.cs
```diff
@@ -33,6 +33,8 @@ public interface IVisualEntity
 
         byte Direction { get; set; }
 
+        byte Size { get; set; }
```
0Lucifer0: Should be IAlive

src/NosCore.GameObject/Services/MapItemBuilder/MapItem.cs
```diff
@@ -66,6 +66,7 @@ public long VisualId
         public short VNum => ItemInstance?.ItemVNum ?? 0;
 
         public byte Direction { get; set; }
+        public byte Size { get; set; }
```
0Lucifer0: No

src/NosCore.Controllers/CharacterScreenPacketController.cs
```diff
@@ -370,6 +370,7 @@ public void SelectCharacter(SelectPacket selectPacket)
                 character.PositionX = character.MapX;
                 character.PositionY = character.MapY;
                 character.Direction = 2;
+                character.Size = 10;
```
0Lucifer0: should be directly in character not on the selection

src/NosCore.GameObject/MapMonster.cs
```diff
@@ -79,6 +80,7 @@ internal void Initialize(NpcMonsterDto npcMonster)
             PositionX = MapX;
             PositionY = MapY;
             Speed = NpcMonster.Speed;
+            Size = 10;
```
0Lucifer0: remove here

src/NosCore.GameObject/MapMonster.cs
```diff
@@ -37,6 +37,7 @@ public class MapMonster : MapMonsterDto, INonPlayableEntity
         public Group Group { get; set; }
         public bool IsSitting { get; set; }
         public byte Speed { get; set; }
+        public byte Size { get; set; }
```
0Lucifer0: Size {get;set;}  = 10;

src/NosCore.GameObject/MapNpc.cs
```diff
@@ -83,6 +84,7 @@ internal void Initialize(NpcMonsterDto npcMonster)
             PositionX = MapX;
             PositionY = MapY;
             Speed = NpcMonster.Speed;
+            Size = 10;
```
0Lucifer0: remove

src/NosCore.GameObject/MapNpc.cs
```diff
@@ -44,6 +44,7 @@ public MapNpc()
         public IDisposable Life { get; private set; }
         public Group Group { get; set; }
         public byte Speed { get; set; }
+        public byte Size { get; set; }
```
0Lucifer0: = 10;

src/NosCore.GameObject/Pet.cs
```diff
@@ -36,6 +36,7 @@ public class Pet : MapMonsterDto, INamedEntity //TODO replace MapMonsterDTO by t
         public DateTime LastMove { get; set; }
         public bool IsSitting { get; set; }
         public byte Speed { get; set; }
+        public byte Size { get; set; }
```
0Lucifer0: = 10;

PR #501, opened 2019-03-13
resolve automaticly Dto and Dao

0Lucifer0:


erwan-joly:
https://stackoverflow.com/questions/43348128/reflection-how-do-i-find-and-invoke-a-local-functon-in-c-sharp-7-0 seems ugly and fun at the same time

PR #514, opened 2019-03-30
Fix Portal, Unfinished fix of groups/party

Adanlink:
Portal now has message for cooldown and respects the 3*3 size of the official.
Groups/parties now works and orders the players like the official (except the sharing of drop of course), but surely my code is improvable and two specifics tests aren't working (and Idk how to fix them because I don't understand them xD), because Idk how in the future will be managed pets this "fix" tries to only do the job with Players.


0Lucifer0:
You shouldn’t merge master on the branch

PR #515, opened 2019-03-30
Fix Portal, Unfinished fix of groups/party

Adanlink:
Portal now has message for cooldown and respects the 3*3 size of the official.
Groups/parties now works and orders the players like the official (except the sharing of drop of course), but surely my code is improvable and two specifics tests aren't working (and Idk how to fix them because I don't understand them xD), because Idk how in the future will be managed pets this "fix" tries to only do the job with Players.


src/NosCore.Controllers/GroupPacketController.cs
```diff
@@ -328,22 +327,27 @@ public void LeaveGroup(PleavePacket __)
 
             if (group.Count > 2)
             {
-                Session.Character.LeaveGroup();
```
0Lucifer0: no

src/NosCore.Controllers/GroupPacketController.cs
```diff
@@ -328,22 +327,27 @@ public void LeaveGroup(PleavePacket __)
...
 
-                    Broadcaster.Instance.Sessions.SendPacket(new InfoPacket
+                    session.SendPacket(new InfoPacket
                     {
                         Message = Language.Instance.GetMessageFromKey(LanguageKey.NEW_LEADER, Session.Account.Language)
-                    }, new EveryoneBut(session.Channel.Id));
+                    });
+                }
+                else
+                {
```
0Lucifer0: no

src/NosCore.Controllers/GroupPacketController.cs
```diff
@@ -328,22 +327,27 @@ public void LeaveGroup(PleavePacket __)
 
             if (group.Count > 2)
             {
-                Session.Character.LeaveGroup();
-
                 if (group.IsGroupLeader(Session.Character.CharacterId))
                 {
+                    Session.Character.LeaveGroup();
```
0Lucifer0: no

src/NosCore.Controllers/GroupPacketController.cs
```diff
@@ -328,22 +327,27 @@ public void LeaveGroup(PleavePacket __)
...
-                        s.VisualId == group.Values.First().Item2.VisualId);
+                        s.VisualId == group.Values.OrderBy(t => t.Item1).FirstOrDefault(u =>
+                        u.Item2.VisualType == VisualType.Player).Item2.VisualId);
 
                     if (session == null)
                     {
                         return;
                     }
 
-                    Broadcaster.Instance.Sessions.SendPacket(new InfoPacket
```
0Lucifer0: everyone in the group is suppose to receive the message

src/NosCore.GameObject/Group.cs
```diff
@@ -59,31 +59,39 @@ public Group(GroupType type)
...
             return new PinitPacket
             {
                 GroupSize = Count == 1 ? 0 : Count,
-                PinitSubPackets = Values.Select(s => s.Item2.GenerateSubPinit(Count == 1 ? i : ++i)).ToList()
+                PinitSubPackets = Values.OrderBy(s => s.Item1).Select(s => s.Item2.GenerateSubPinit(Count == 1 ? i : ++i)).ToList()
             };
         }
 
-        public List<PstPacket> GeneratePst()
+        public List<PstPacket> GeneratePst(long visualId)
```
0Lucifer0: no

src/NosCore.GameObject/Group.cs
```diff
@@ -59,31 +59,39 @@ public Group(GroupType type)
...
         {
-            var i = 0;
+            var i = 1;
+
+            INamedEntity avoid(INamedEntity namedEntity)
```
0Lucifer0: no

src/NosCore.Controllers/GroupPacketController.cs
Adanlink: That's not how works official, but then we need to change the message to something like: "The party master is [{0}]".

src/NosCore.Packets/ServerPackets/PinitPacket.cs
```diff
@@ -28,7 +28,7 @@ public class PinitPacket : PacketDefinition
         [PacketIndex(0)]
         public int GroupSize { get; set; }
 
-        [PacketIndex(1, SpecialSeparator = "|")]
+        [PacketIndex(1, IsOptional = true, SpecialSeparator = "|")]
```
0Lucifer0: i don't think this is optional

src/NosCore.GameObject/Group.cs
Adanlink: Then how we avoid sending useless info to the same player who "requests" the status of his party? (official avoids sending this info)

src/NosCore.Packets/ServerPackets/PinitPacket.cs
Adanlink: Is optional (I think). When there is only the player without any party the packet sent needs to be "pinit 0", not "pinit 1 1|1|1|etc"

src/NosCore.Controllers/GroupPacketController.cs
0Lucifer0: are you sure ?

src/NosCore.GameObject/Group.cs
0Lucifer0: what do you mean ? the player need to know who is in his party how could it work without ?

src/NosCore.Controllers/GroupPacketController.cs
Adanlink: If the player leaves why we check later if he/she is the groupleader? It will never detect him as the groupleader and never pass the if

src/NosCore.Packets/ServerPackets/PinitPacket.cs
0Lucifer0: right... actually it should be an empty list as 0 is the count

src/NosCore.Controllers/GroupPacketController.cs
Adanlink: xD, I have coded this today basing it in the official US-Cylloan (using 3 acc at the same time and making sure all the packets are the same ;) )

src/NosCore.GameObject/Group.cs
Adanlink: Sorry for explaining incorrectly. We need to know who is requesting the info because we need to avoid sending his own info. Player A doesn't need to know his own information via "pst" packet, but needs to know Player B and C. This is respected by official.

0Lucifer0: the you need a better name than visualId and also use the VisualEntity instead of the id

Adanlink: Okay, will do it when I can!

src/NosCore.GameObject/Group.cs
Adanlink: Any other method to do this? xD, sorry for being a noob

0Lucifer0: just keep the code as it was and just where != namedEntity

Adanlink: With as it was you mean as I done it or how it had been before I touched it?. I mean, the code that you want to keep used to update the "grouporder" while making the PstPacket. This means that if I put a ".where" and remove the "avoid" that will cause the order to mess, messing up the info NosTale gets.

0Lucifer0: You avoid is just doing a ++i which was already done in the code

src/NosCore.Controllers/GroupPacketController.cs
0Lucifer0: Because we check before not after line 333

src/NosCore.GameObject/Group.cs
Adanlink: But if we use ".where" it will only do "++i" when not sending the info. Let me explain. Grouporder needs to prevail even if you don't send that "pst", so, doing "++i" only when sending that exact "pst" will mess up the order

src/NosCore.Controllers/GroupPacketController.cs
Adanlink: Mmm... I don't get it. How we checked before if the code was "LeaveGroup(Player A)" and then "Check if Player A is the leader of the group"?

src/NosCore.GameObject/Group.cs
0Lucifer0: So you just want the order not i

Adanlink: This are the order for pinit and pst (If I don't remember badly) -> 0 = companion | 1 = pet | 2 = Party Master | 3 = Party member (next party master if Party Master leaves the party) | 4 = Party member. I use "i" as the grouporder (because it was used too in the original code) and the "lastId" as the Order

PR #577, opened 2019-05-30
Optimise parser

0Lucifer0:


src/NosCore.Database/DAL/ItemInstanceDAO.cs
```diff
@@ -222,52 +222,42 @@ public SaveResult InsertOrUpdate(IEnumerable<IItemInstanceDto> dtos)
...
-                                    entity.Adapt<ItemInstanceDto>().Adapt<ItemInstance>();
+                        : dto.Adapt<ItemInstance>(), _primaryKey.GetValue(dto, null)));
+                }
+                var ids = list.Select(s => s.Item2).ToArray();
+                var dbkey = typeof(ItemInstance).GetProperty(_primaryKey.Name);
+                var entityfounds = dbset.FindAllAsync(dbkey, ids).ToList();
 
+                foreach (var dto in list)
+                {
+                    var entity = dto.Item1 is BoxInstance ? dto.Item1.Adapt<BoxInstanceDto>().Adapt<BoxInstance>() : dto.Item1 is SpecialistInstance
```
Elendan: Not very important here, but such a ternary is not really readable, I would rather use proper if's even if statements or a switch instead of a 5 lines ternary condition

erwan-joly: just move the old one :) this whole file is not really a good thing but required

PR #579, opened 2019-06-01
Fix friend system

0Lucifer0:


0Lucifer0:
will again merge and fix later as pr too long

PR #595, opened 2019-06-06
added PL keys, MsgPacket in ChangeClass

imquarry:


0Lucifer0:
Character is not suppose to change

0Lucifer0:
Also you don’t want all those space change it make it impossible to review

0Lucifer0:
Change class is not suppose to have a msg as it’s not an official behavior here

PR #615, opened 2019-06-30
start adding new auth system

0Lucifer0:


0Lucifer0:
will add test and fix codacy and this will be done :)

PR #633, opened 2019-07-17
Player name in exchange

yoshwx:
#600


0Lucifer0:
Test fail cause of bad number of parameters in existing lang.please fix them by adding {0}

PR #808, opened 2019-11-16
Translation enhancements

emil-sergeev:


0Lucifer0:
Error Message:
   Assert.Fail failed. MINILAND_VISITOR does not contain the correct amount of parameters. Expected:1 Given: 2NEW_MAIL does not contain the correct amount of parameters. Expected:1 Given: 0

emil-sergeev:
image

Hm. Okay, I agree that NEW_MAIL doesn't contain a required param. But why MINILAND_VISITOR have only 1 instead 2 params?

image

0Lucifer0:
no idea i would say that there is a bug in the regexp. will have a look. fix the first one then i will just merge and fix this later

PR #840, opened 2019-11-23
Added lot of missing czech translate

Goblinek:


0Lucifer0:
The file is not suppose to be in this directory but in data/resource

PR #841, opened 2019-11-24
Add lot of missing czech translate

Goblinek:


0Lucifer0:
Test failing and file format is incorrect

PR #947, opened 2020-02-23
fix duplicated foreign keys

Zagith:
some tables has duplicated foreign keys and fields (like NpcMonsterVNum and SkillVNum)


0Lucifer0:
Doesn’t look like duplicate but missing one

PR #1024, opened 2020-04-12
serhocanszd Traslation TR 

serhocanszd:
Translation for TR , LocalizedResources.Tr


erwan-joly:
remove all those delete commits

PR #1027, opened 2020-04-12
[serodyx] Translation [TR]

serodyx:
added some values


erwan-joly:
NPCMONSTERTALKS_LOADED  doesn't exist

PR #1617, opened 2022-01-04
Example with tests

0Lucifer0:


0Lucifer0:
https://github.com/NosCoreIO/NosCore/pull/1618 don't need the toString

PR #1631, opened 2022-01-07
Fix mobs xp, jxp and hxp calculation

Fizo55:


src/NosCore.Parser/Parsers/NpcMonsterParser.cs
```diff
@@ -112,7 +111,9 @@ public async Task InsertNpcMonstersAsync(string folder)
                 {nameof(NpcMonsterDto.DarkResistance), chunk => Convert.ToInt16(chunk["ATTRIB"][0][7])},
                 {nameof(NpcMonsterDto.MaxHp), chunk => Convert.ToInt32(chunk["HP/MP"][0][2]) + _basicHp[Level(chunk)]},
                 {nameof(NpcMonsterDto.MaxMp), chunk => Convert.ToInt32(chunk["HP/MP"][0][3]) + Convert.ToByte(chunk["RACE"][0][2]) == 0 ? _basicPrimaryMp[Level(chunk)] : _basicSecondaryMp[Level(chunk)]},
-                {nameof(NpcMonsterDto.Xp), chunk =>  Math.Abs(Convert.ToInt32(chunk["EXP"][0][2]) + _basicXp[Level(chunk)])},
+                {nameof(NpcMonsterDto.Xp), chunk =>  Level(chunk) >= 19 ?
```
0Lucifer0, line 114: Please make that a function

Fizo55: Already done in local, I'm fixing jxp aswell that's why it's not pushed rn

src/NosCore.Database/Entities/NpcMonster.cs
```diff
@@ -87,8 +87,6 @@ public NpcMonster()
 
         public byte HeroLevel { get; set; }
 
-        public int HeroXp { get; set; }
```
Fizo55, line 90: On official, there is no data about that in the monster informations, it's working like that :
If a monster is in act 6 - 7 so he can give u some hxp (monsterXp / 25)

erwan-joly: this is still calculation that should happen on parser time to make it more efficient

erwan-joly: just set HXP to monsterXp/25

PR #2058, opened 2026-04-19
Finish ECS player migration and fix map change bugs

erwan-joly:
## Summary
- Preserves **all** 17 player components across map transitions via a new `MapWorld.ClonePlayer` helper. Previously only `PlayerStateComponent` survived, silently wiping Invisible, Morph, combat stats, SP, flags, timing, etc. on every teleport/map change.
- Fixes the `NullReferenceException` that surfaced on subsequent map changes: `Group` boxed the original `PlayerComponentBundle` at character select; after the entity was destroyed and recreated, the stale box hit `World.Has<T>(destroyedEntity)` on every `GeneratePst`/`GeneratePidx` call. We now re-register the character in its group after `SetPlayerEntity`.
- `ChangeMapAsync` was passing `session.Character.MapInstanceId` (read from `PositionComponent`, not yet updated) to `ChangeMapInstanceAsync`, so teleports with an explicit `mapId` actually stayed on the current map. Fixed by using `mapInstance.MapInstanceId` directly.
- Miniland was never initialized on login (registered service, never called). Added the call in `SelectPacketHandler` and switched `IMinilandService.InitializeAsync` + `ICharacterInitializationService.InitializeAsync` from the legacy `Character` class to `PlayerComponentBundle`. Replaced `Miniland.CharacterEntity` (a stale boxed bundle) with `CharacterName`, and switched `MinilandEntranceHandler` to compare ownership by `OwnerId`.
- Initial player `Size` set to `100` in `MapWorld.CreatePlayer` (was `10`), matching the fix from commit 41d86b32 that was lost in the ECS migration.

## Test plan
- [ ] Login, select character, walk around starting map
- [ ] Teleport / change map via command; verify HP/MP, Morph, Invisible, combat state, SP, flags persist
- [ ] Group: form a group, change map — other members still visible, no NRE
- [ ] Miniland: enter own miniland, verify `GetMinilandPortals`/`MinilandEntranceHandler` works
- [ ] Character appears at normal size (not tiny) on arrival


erwan-joly:
Closing — branch has diverged significantly from master (master shipped overlapping ECS work via PRs #2060-#2064). Will reopen as a fresh PR after rebasing the remaining novel work (facade deletion + interface removal).

PR #2278, opened 2026-08-22
fix: WorldServer fails to start, both under Wolverine 6 and with redi…

denislauri1999:
…rected output

Two failures no build can see. Both were found by actually starting the servers.

1. WOLVERINE 6 MOVED THE RUNTIME COMPILER OUT OF THE MAIN PACKAGE

   The dependency bump took WolverineFx from 5.37 to 6.28. From 6 on, the runtime compiler lives in WolverineFx.RuntimeCompilation. The server builds without a single warning and then dies on startup with:

       no IAssemblyGenerator (Roslyn) is registered

   Handlers are generated on first run, and there is nobody left to compile them.

   Adding the package reference is not enough, and this is the part worth knowing: WolverineFx.RuntimeCompilation registers itself AS A WOLVERINE EXTENSION, and UseNosCoreWolverine passes ExtensionDiscovery.ManualOnly - which is precisely what turns extension discovery off. It has to be invoked by hand with opts.UseRuntimeCompilation().

2. THE CONSOLE TITLE IS TOUCHED WITHOUT CHECKING WHETHER THERE IS A CONSOLE

   MasterServer, LoginServer and WorldServer all set Console.Title and print a banner without checking Console.IsOutputRedirected. On Windows, with output redirected - a service, a container, a process manager, CI - touching the title throws, and the process dies during startup having done nothing. All that is left in the log is one line: "Invalid handle".

   The Parser already had this guard; the other three did not.

   This means NosCore currently cannot be run headless at all, which matters for anyone hosting it rather than running it from a terminal. Both the title and the banner are decoration: they are attempted, and skipped when they cannot be done.

VERIFIED by starting Master, Login and World with output redirected to files. All three come up, the World loads 777 maps and 7697 items from the database, registers with the master, and the scheduler announces its events.

Build clean at 0 warnings in Debug and Release.

NOTE, unrelated to this change: NosCore.Core.Tests.CheckLanguageUsage fails under `dotnet test -c Release` and passes under Debug, on master as well as here. Worth a look on its own - anyone validating a contribution with a Release build will see a red suite and reasonably assume the contribution caused it.

## Summary by CodeRabbit

* **New Features**
  * Added runtime compilation support for improved startup and handler discovery.

* **Bug Fixes**
  * Prevented login, master, and world servers from failing when console output is redirected or unavailable.
  * Made startup banners and Windows console title updates safely ignore console-related errors.

### DIFF
--- a/src/NosCore.GameObject/Messaging/WolverineHostExtensions.cs
+++ b/src/NosCore.GameObject/Messaging/WolverineHostExtensions.cs
@@ -1,4 +1,4 @@
-﻿//  __  _  __    __   ___ __  ___ ___
+//  __  _  __    __   ___ __  ___ ___
 // |  \| |/__\ /' _/ / _//__\| _ \ __|
 // | | ' | \/ |`._`.| \_| \/ | v / _|
 // |_|\__|\__/ |___/ \__/\__/|_|_\___|
@@ -29,14 +29,8 @@ namespace NosCore.GameObject.Messaging
                 opts.ServiceName = serviceName;
                 opts.ApplicationAssembly = typeof(WolverineHostExtensions).Assembly;
 
-                // From Wolverine 6 on, the runtime compiler lives in a separate package
-                // (WolverineFx.RuntimeCompilation) which normally registers itself as an
-                // extension. Referencing it is not enough here: ExtensionDiscovery.ManualOnly
-                // below turns extension discovery off, so it has to be invoked by hand.
-                //
-                // Without this the server builds clean and then dies on startup with "no
-                // IAssemblyGenerator (Roslyn) is registered" - handlers are generated on first
-                // run and there is nobody left to compile them. No build can catch it.
+                // ExtensionDiscovery.ManualOnly below stops WolverineFx.RuntimeCompilation from
+                // self-registering, so it has to be invoked here.
                 opts.UseRuntimeCompilation();
                 foreach (var asm in handlerAssemblies)
                 {
--- a/src/NosCore.GameObject/Messaging/WolverineHostExtensions.cs
+++ b/src/NosCore.GameObject/Messaging/WolverineHostExtensions.cs
@@ -1,4 +1,4 @@
-//  __  _  __    __   ___ __  ___ ___
+﻿//  __  _  __    __   ___ __  ___ ___
 // |  \| |/__\ /' _/ / _//__\| _ \ __|
 // | | ' | \/ |`._`.| \_| \/ | v / _|
 // |_|\__|\__/ |___/ \__/\__/|_|_\___|
@@ -28,6 +28,16 @@ namespace NosCore.GameObject.Messaging
             {
                 opts.ServiceName = serviceName;
                 opts.ApplicationAssembly = typeof(WolverineHostExtensions).Assembly;
+
+                // From Wolverine 6 on, the runtime compiler lives in a separate package
+                // (WolverineFx.RuntimeCompilation) which normally registers itself as an
+                // extension. Referencing it is not enough here: ExtensionDiscovery.ManualOnly
+                // below turns extension discovery off, so it has to be invoked by hand.
+                //
+                // Without this the server builds clean and then dies on startup with "no
+                // IAssemblyGenerator (Roslyn) is registered" - handlers are generated on first
+                // run and there is nobody left to compile them. No build can catch it.
+                opts.UseRuntimeCompilation();
                 foreach (var asm in handlerAssemblies)
                 {
                     opts.Discovery.IncludeAssembly(asm);
--- a/src/NosCore.GameObject/NosCore.GameObject.csproj
+++ b/src/NosCore.GameObject/NosCore.GameObject.csproj
@@ -1,4 +1,4 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
@@ -20,6 +20,9 @@
     <PackageReference Include="NosCore.Algorithm" />
     <PackageReference Include="System.Reactive" />
     <PackageReference Include="WolverineFx" />
+    <!-- See the note in Directory.Packages.props: from Wolverine 6 on, without this the
+         server builds but will not start. -->
+    <PackageReference Include="WolverineFx.RuntimeCompilation" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/NosCore.GameObject/NosCore.GameObject.csproj
+++ b/src/NosCore.GameObject/NosCore.GameObject.csproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>net10.0</TargetFramework>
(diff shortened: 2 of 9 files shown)


src/NosCore.GameObject/Messaging/WolverineHostExtensions.cs
```diff
@@ -28,6 +28,16 @@ public static IHostBuilder UseNosCoreWolverine(this IHostBuilder builder, string
             {
                 opts.ServiceName = serviceName;
                 opts.ApplicationAssembly = typeof(WolverineHostExtensions).Assembly;
+
+                // From Wolverine 6 on, the runtime compiler lives in a separate package
```
erwan-joly: don't need all those comments

src/NosCore.GameObject/NosCore.GameObject.csproj
```diff
@@ -20,6 +20,9 @@
     <PackageReference Include="NosCore.Algorithm" />
     <PackageReference Include="System.Reactive" />
     <PackageReference Include="WolverineFx" />
+    <!-- See the note in Directory.Packages.props: from Wolverine 6 on, without this the
```
erwan-joly: no need for  that comments

src/NosCore.LoginServer/LoginServer.cs
```diff
@@ -32,7 +32,9 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
         {
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                Console.Title += $@" - Port : {Convert.ToInt32(loginConfiguration.Value.Port)}";
+                // The console title cannot be touched when output is redirected.
+                try { Console.Title += $@" - Port : {Convert.ToInt32(loginConfiguration.Value.Port)}"; }
+                catch { /* console rediretta o assente */ }
```
erwan-joly: comments lang is wrong

src/NosCore.LoginServer/LoginServerBootstrap.cs
```diff
@@ -75,7 +75,9 @@ private static void InitializeConfiguration(string[] args, IServiceCollection se
             services.AddOptions<ServerConfiguration>().Bind(conf).ValidateDataAnnotations();
             services.AddOptions<WebApiConfiguration>().Bind(conf.GetSection(nameof(LoginConfiguration.MasterCommunication))).ValidateDataAnnotations();
 
-            Logger.PrintHeader(ConsoleText);
+            // The banner measures the window width, which throws when output is redirected.
+            // It is decoration, not a reason to refuse to start.
+            try { Logger.PrintHeader(ConsoleText); } catch { /* console rediretta o assente */ }
```
erwan-joly: ditto

src/NosCore.LoginServer/LoginServerBootstrap.cs
```diff
@@ -75,7 +75,9 @@ private static void InitializeConfiguration(string[] args, IServiceCollection se
             services.AddOptions<ServerConfiguration>().Bind(conf).ValidateDataAnnotations();
             services.AddOptions<WebApiConfiguration>().Bind(conf.GetSection(nameof(LoginConfiguration.MasterCommunication))).ValidateDataAnnotations();
 
-            Logger.PrintHeader(ConsoleText);
+            // The banner measures the window width, which throws when output is redirected.
```
erwan-joly: no need for that comment and likely it can be baked in the PrintHeader  instead

src/NosCore.LoginServer/LoginServerBootstrap.cs
```diff
@@ -166,7 +168,7 @@ private static IHost BuildHost(string[] args)
                 {
                     if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
                     {
-                        Console.Title = Title;
+                        try { Console.Title = Title; } catch { /* console rediretta o assente */ }
```
erwan-joly: ditto

src/NosCore.MasterServer/MasterServer.cs
```diff
@@ -27,7 +27,9 @@ protected override Task ExecuteAsync(CancellationToken stoppingToken)
             logger.LogInformation(logLanguage[LogLanguageKey.SUCCESSFULLY_LOADED]);
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                Console.Title += $@" - WebApi : {_masterConfiguration.WebApi}";
+                // See MasterServerBootstrap: this throws when output is redirected.
+                try { Console.Title += $@" - WebApi : {_masterConfiguration.WebApi}"; }
+                catch { /* console rediretta o assente */ }
```
erwan-joly: ditto

src/NosCore.MasterServer/MasterServerBootstrap.cs
```diff
@@ -88,11 +88,17 @@ private static WebApplication BuildApp(string[] args)
...
-            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            // When output is redirected - a Windows service, a container, a process manager,
+            // CI - touching the console title throws on Windows, and the server dies during
+            // startup having done nothing at all. The banner has the same problem: it measures
+            // the window width. Both are decoration, so they are attempted and skipped when
+            // they cannot be done.
+            if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows) && !Console.IsOutputRedirected)
             {
-                Console.Title = Title;
+                try { Console.Title = Title; } catch { /* console rediretta o assente */ }
```
erwan-joly: ditto

src/NosCore.WorldServer/WorldServer.cs
```diff
@@ -49,7 +49,9 @@ protected override async Task ExecuteAsync(CancellationToken stoppingToken)
 
             if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
             {
-                Console.Title += $@" - Port : {worldConfiguration.Value.Port}";
+                // The console title cannot be touched when output is redirected.
+                try { Console.Title += $@" - Port : {worldConfiguration.Value.Port}"; }
```
erwan-joly: ditto

denislauri1999:
All addressed, and the PR is now net -14 lines.

**Comments:** removed. Only two survive: one line on `opts.UseRuntimeCompilation()` saying why it has to be called by hand, and the summary on the new helper.

**Language:** you were right — some `catch` blocks had Italian in them. Gone.

**"can be baked in the PrintHeader instead":** taken further than that. There is now a `ConsoleDecoration` helper in `NosCore.Core` with `SetTitle`, `AppendToTitle` and `PrintHeader`, each guarded once, and every call site is a plain one-liner with no try/catch and no platform check. That also picked up two sites the first version had missed — including the parser's, which was still unguarded.

I had also introduced BOMs on the file headers; reverted, so the diff no longer shows those lines as changed.

denislauri1999:
Ignore the comment above — I wrote it before fetching and did not see you had already pushed `d5123ffa`. Your version is better than what I had: guarding on `Console.IsOutputRedirected` beats swallowing the exception, and dropping the `PrintHeader` try/catch is right if Shared 6.0.1 handles it internally. I have thrown away my `ConsoleDecoration` helper and taken yours.

One thing my version had picked up that yours may not: `ParserBootstrap` sets the console title too, and it was unguarded before this PR. Worth checking it got the same treatment.

Sorry for the near-miss on the force-push — `--force-with-lease` caught it. I will fetch before touching any of these branches from now on.

PR #2279, opened 2026-08-22
fix(parser): the sign of an item BCard's first value selects the subtype

denislauri1999:
WHAT. Every BCard subtype comes in a pair the official files describe as opposites, and the sign of the first value says which half is meant:

    type 11, subtype 11:  "damage taken from all attacks is INCREASED by %s%%"
    type 11, subtype 12:  "damage from all attacks is REDUCED by %s%%"

A negative first value means the second of the pair. SkillParser has always applied this, CardParser has always applied it, ItemParser never did.

WHAT IT COST. All 132 type-11 effects on equipment in the game files carry a negative first value, because armour reduces the damage you take - that is what armour is. Stored without the sign they all became subtype 11. The combat code looks up subtype 12, finds nothing, and the reduction silently never applies. Every piece of armour in the game has been weaker than it reads, with nothing logged.

Same for type 45 (48 effects, all negative: debuff resistance), type 114 (44 of 50) and type 44 (309 of 763, which are defence rather than attack). Around 530 effects in total.

VERIFIED THREE WAYS, because "the other two parsers do it" is not on its own proof:

  1. the two sibling parsers apply the rule to the same file family;
  2. this parser ALREADY treats a negative first value specially, taking its magnitude and dividing by four - which only makes sense if the sign carries meaning;
  3. the rendered game text matches. Item 119 "Spirit Tunic" declares first = -40, and the client shows "There is a 10% chance that damage from all attacks is reduced by 10%" - the reduced variant, with 40/4 = 10 for the chance.

CONFIRMED IN THE DATABASE after re-parsing: every type-11 effect on an item is now subtype 12, 22, 32 or 42 - the "reduced" halves - and item 119 reads exactly as the client renders it.

Three tests, written on the real items: 119 Spirit Tunic (negative, all attacks), a positive counterpart, and 117 Nubuck Tunic (negative, melee, subtype index 1 -> 22).

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected BCard effect variant selection when negative values are provided.
  * Ensured the appropriate alternate effect subtype is selected for negative values.

* **Tests**
  * Updated BCard subtype test coverage to reflect the corrected behavior.

### DIFF
--- a/src/NosCore.Parser/Parsers/ItemParser.cs
+++ b/src/NosCore.Parser/Parsers/ItemParser.cs
@@ -1,4 +1,4 @@
-﻿//  __  _  __    __   ___ __  ___ ___
+//  __  _  __    __   ___ __  ___ ___
 // |  \| |/__\ /' _/ / _//__\| _ \ __|
 // | | ' | \/ |`._`.| \_| \/ | v / _|
 // |_|\__|\__/ |___/ \__/\__/|_|_\___|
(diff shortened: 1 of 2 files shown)


src/NosCore.Parser/Parsers/ItemParser.cs
```diff
@@ -197,6 +197,36 @@ public async Task ParseAsync(string folder)
...
 
 
+        // Effects declared by an item, from the BUFF section: five groups of five fields, laid
+        // out as [type, first, second, subtype index, third].
+        //
+        // THE SIGN OF `first` SELECTS THE VARIANT, and this parser was the only one of the three
+        // that did not honour it. Every BCard subtype comes in pairs which the official files
+        // describe as opposites:
+        //
+        //     type 11, subtype 11:  "damage taken from all attacks is INCREASED by %s%%"
```
erwan-joly: i don't think we need this comment

PR #2280, opened 2026-08-22
fix(data): BCard type 56 and 57 are swapped, and eleven enums describ…

denislauri1999:
…e the wrong effect

`BCard.dat` is a dictionary: for every effect type it carries a name and five subtype pairs, each resolving to readable English in the language files. The enums here were derived rather than read, and they have drifted from it.

None of this raises anything. A wrong number sends an effect into another type's handler; a wrong name makes an effect look handled while it does something else. Both produce a working game with the wrong behaviour, which is the only kind of bug that matters in this domain.

--- TWO TYPES ARE SWAPPED ---

The file declares 56 as "4th Glacernon Skill" and 57 as "Attack of the summoned monster". The enum had them the other way round.

--- MEMBERS THAT ARE NOT VALID SUBTYPES ---

Types 93 and 98 each had two members numbered 00 and 01. Every subtype in this format is (index+1)*10+1 or +2, so those numbers cannot occur; anything switching on them was dead.

--- NAMES THAT DESCRIBE A DIFFERENT EFFECT ---

    98/11-12   was "AllAttackDamageRecived"; the file says "all elemental damage is
               increased by %s%%" - the other side of the hit
    107/31-32  was "XpGain"; the file says "The maximum additional HP increases by %s%%"
    107/51-52  was a dodge pair; the file gives 51 as dodge and 52 as HIT RATE, both
               increased. Calling 52 "DecreaseDodge" takes dodge away where the file
               grants hit rate
    16/31      was numbered 41, the same as another member, so its branch was unreachable

--- SUBTYPES MARKED AS NOT EXISTING, WHICH DO ---

Type 109's first three pairs were commented "didn't exist"; they are the lion, eagle and snake loas, declared ten times each on Cards. Type 95 declared three of its ten members and said subtype 31 "didn't exist" - it appears twelve times in Card.dat alone. Type 93 marked light and shadow the same way, with a question mark.

--- AND ENUMS THAT STOPPED EARLY ---

Types 47, 49, 53, 62, 78, 86, 104, 108 and SecondSpCard carried between two and eight of their ten members. Filled in from the files, with the sentence each one comes from in the doc comment so the next reader does not have to look it up.

Where a pair's two halves carry the SAME sentence in the files - types 1, 2, 86/11-12 - that is noted, because the inherited "...Negated" naming reads as if the second half undoes the first and it does not.

Two files, no behaviour changed, 0 warnings, all tests pass.

These were found by lining the enum's names up against the ones BCard.dat declares, rather than checking them one at a time. Happy to share that script if it would be useful.

## Summary by CodeRabbit

* **Enhancements**
  * Expanded and corrected buff, debuff, elemental, arena, summoned-monster, and card effect definitions.
  * Added support for additional effect subtypes, including damage, resistance, penalties, upgrades, and resource-related effects.
  * Added new card types for Type 114, armour piercing, heat and gravity, fuel points, and token systems.
  * Corrected card type assignments and clarified effect names and descriptions for more accurate game data.


src/NosCore.Data/Enumerations/Buff/AdditionalTypes.cs
```diff
@@ -34,6 +34,18 @@ public enum Absorption : byte
             MagicalAttacksDecreased = 42
         }
 
+        /// <summary>
```
erwan-joly: far too many useless comments and some are not in  english

denislauri1999:
Thanks — both fair, and both fixed.

**Comments:** cut all 230 of them. The file now carries the enum members and nothing else. Same pass is going over the other five PRs.

**Language:** you were right, a few Italian lines had slipped in, along with references to tooling that only exists in my own tree. Gone.

Also added Type 114 subtypes 51-52 (`Reduces enemies' soft crit damage by %s%%`), which I had missed — CodeRabbit caught that one.

It also flagged `RestoreMpFromDamageTaken` (53/32) and `RestoreMpFromDamage` (47/22) as misnamed HP effects. I checked `BCard.dat` and left them: it declares 53/32 as `MP is increased by %s%% of damage taken` and 47/22 as `Restores MP equal to %s%% of the damage inflicted`, so the MP in those names is the file's.

PR #2281, opened 2026-08-22
feat(mate): read back the mates the capture service has been writing

denislauri1999:
WHAT: a Mate game object, a MateService that loads a character's mates, a PlayerMatesComponent on the player bundle, loading at character select, and the pet-list burst at game start.

WHY: CaptureService writes a MateDto row every time a pet is caught, and nothing ever read it. The row went into the database and the pet was never heard from again — the catch worked and produced nothing a player could see.

SOURCE: a real packet capture. It settles three things that would otherwise have been guesses:

  * the login order — p_clear, one sc_p per pet and sc_n per partner, then sc_p_stc;
  * that pets and partners are numbered from zero SEPARATELY (sc_p slots 0..7 sit next to sc_n slots 0..1 in the same burst);
  * the experience table. The curve the older emulators ship is twenty times the pet requirement and five times the partner one — ten observations from level 1 to level 88 match after dividing, two of them eight-digit numbers, so it is not curve-fitting. MateXpTable carries the divisors and all ten observations are tests. Worth a look on its own: nothing throws when this is wrong, the pet just never levels, and it reads as grind rather than as a bug.

EXPECTED: a caught pet survives the session it was caught in. Log in and the pet window lists what you own, with its level, loyalty and experience bar.

OUT OF SCOPE, each for a reason written next to it in the code:

  * the per-level HP/MP and damage curves. The inherited ones reproduce no captured row — a level-3 chicken would get 262 HP where the server sent 195 — so the mate reports the creature's declared statistics: exact at level 1, low above it. Sixteen samples across three stat families are not enough to derive the right curve; fitting one would be inventing it.
  * pinit for mates. The capture gives a mate row eight fields; PinitSubPacket serialises eleven, because its last three members are non-nullable value types. Sending a shape an authoritative source contradicts is worse than not sending it — happy to send a NosCore.Packets change for this if you want it.
  * empty partner equipment slots. The capture writes them as a bare -1; leaving the sub-packet null makes the serialiser drop the separating space and emit "1536-1-1-1". All three numbers are filled instead (-1.0.0), which keeps the field count right. Unreachable today, since nothing creates a partner yet.
  * summoning onto the map, combat, loyalty, feeding, the partner's specialist card. Each is its own slice.

## Summary by CodeRabbit

* **New Features**
  * Added support for loading, saving, and displaying pets and partners.
  * Pets and partners persist across map changes and character selection.
  * Game startup now synchronizes mate information, including names, slots, health, and experience.
  * Pets and partners follow players and are positioned on nearby walkable spaces.
  * Added mate attacks against valid targets.
  * Captured pets are automatically assigned as active when no other pet is active.

* **Tests**
  * Added coverage for mate persistence, placement, packets, naming, following, attacks, and experience requirements.


src/NosCore.GameObject/Messaging/WolverineDependencyRegistrar.cs
```diff
@@ -47,6 +48,9 @@ public static void RegisterDependencies(IServiceCollection services)
         services.AddSingleton<IIdService<Group>>(_ => new IdService<Group>(1));
         services.AddSingleton<IIdService<MapItemComponentBundle>>(_ => new IdService<MapItemComponentBundle>(100000));
         services.AddSingleton<IIdService<ChannelInfo>>(_ => new IdService<ChannelInfo>(1));
+        // Mates start at two million so their transport ids cannot collide with the visual ids
+        // of the monsters and npcs already on a map, which live far below that.
+        services.AddSingleton<IIdService<Mate>>(_ => new IdService<Mate>(2000000));
```
erwan-joly, line 51: did you check that on  official ?

src/NosCore.GameObject/Services/MapChangeService/MapChangeService.cs
```diff
@@ -123,6 +123,9 @@ public async Task ChangeMapInstanceAsync(ClientSession session, Guid mapInstance
                 var inventory = oldWorld.TryGetComponent<PlayerInventoryComponent>(oldEntity) ?? default;
                 var social = oldWorld.TryGetComponent<PlayerSocialComponent>(oldEntity) ?? default;
                 var requests = oldWorld.TryGetComponent<PlayerRequestsComponent>(oldEntity) ?? default;
+                // The mates come across untouched: a map change moves where the character is, not
```
erwan-joly: no need for that comment

src/NosCore.GameObject/Services/MateService/Mate.cs
```diff
@@ -0,0 +1,201 @@
...
+        {
+            return new ScnPacket
+            {
+                PetId = PetSlot,
+                NpcMonsterVNum = VNum,
+                TransportId = MateTransportId,
+                Level = Level,
+                Loyalty = Loyalty,
+                Experience = Experience,
+                // A partner's four equipment slots. Nothing wears anything yet.
```
erwan-joly: no need for that comment

src/NosCore.GameObject/Services/MateService/Mate.cs
```diff
@@ -0,0 +1,201 @@
...
+            if (string.IsNullOrEmpty(name))
+            {
+                // A creature with no entry for that language would otherwise be nameless; EN is
+                // what the parser always fills, so it is the only safe fallback.
+                name = NpcMonster.Name.TryGetValue(language, out var localized)
+                    ? localized
+                    : NpcMonster.Name[RegionType.EN];
+            }
+
+            return name.Replace(' ', '^');
```
erwan-joly: that should be sorted by the packet serialization itself not manually

src/NosCore.GameObject/Services/MateService/MateXpTable.cs
```diff
@@ -0,0 +1,111 @@
...
+// | | ' | \/ |`._`.| \_| \/ | v / _|
+// |_|\__|\__/ |___/ \__/\__/|_|_\___|
+//
+
+using NosCore.Data.Enumerations.Character;
+
+namespace NosCore.GameObject.Services.MateService
+{
+    /// <summary>
+    /// How much experience a pet or a partner needs to reach the next level.
```
erwan-joly: this has nothing to do here we have a library for stats/xp calculation

denislauri1999:
Comments cut — 179 lines gone.

**"we have a library for stats/xp calculation":** agreed. Opened NosCoreIO/NosCore.Algorithm#129 with `MateExperienceService` (and `FamilyExperienceService` for the same complaint on #2283), following the shape of the existing services and with approval tables under `documentation/`. I have left `MateXpTable` in this PR only so it stays buildable — it goes the moment that package ships, and I am happy to hold this PR until then if you would rather not merge the duplicate.

**"that should be sorted by the packet serialization itself not manually":** you are right, the space-to-caret substitution has no business being at every call site. It belongs in the serializer, alongside a second problem I hit: a null sub-packet loses its separating space, so `sc_n` comes out as `1536-1-1-1` and `gidx` as `gidx 1 741328-1 - 0`, neither of which the client can split. I will send both as one PR to NosCore.Packets and strip the manual `Replace` here once it lands.

**"did you check that on official?"** (the 2 000 000 id seed) — no, and thank you for asking, because the honest answer is that I took it from the older emulator's convention rather than from evidence. The capture shows official mate transport ids in the same five-digit range as everything else (26716-26724), so the real server does not keep a separate band at all. Ours only has to avoid colliding with monster and npc visual ids, which start well below that — but if you would rather mates draw from the same space as other entities, say so and I will change it.

denislauri1999:
Follow-up on the escaping: you were right, and it needed no library change at all — `StringSerializer` already replaces the separator with a caret for every non-final string field, so my `Replace` was doing it twice. Removed here and on #2283.

The null sub-packet problem was real though, and is up as NosCoreIO/NosCore.Packets#493.

denislauri1999:
Extended this one rather than opening a second PR, since it is the same feature: **the mate now actually appears on the map.**

Before this it loaded and showed in the pet window and that was all — nothing put it anywhere you could see it. It now spawns beside its owner on arrival, is announced to everyone else on the map, gets its health bar, and is taken away when the owner leaves.

The shapes come off the capture, not from the older emulator:

```
in 2 1506 445562 26 26 2 100 100 0 0 3 626114 1 0 -1 Ratufu^pirate^(Feu) 0 -1 ...
pst 2 22687 0 100 100 24471 3100 0 0 0
```

`Owner` and `GroupEffect=3` are what separate a mate from a map npc — without them the client draws it as scenery. The byte after the name is 1 for both partners in the capture and 0 for every pet. And `pst`'s third field carries the mate type where a player's carries a party position.

A freshly caught pet joins the team if the pet slot is free, because that is what catching one does in game — it walks out beside you rather than going into storage.

**Not in this PR:** the mate stands where it is put. Following the owner step by step wants the movement system and is its own change.

denislauri1999:
Three more from review, all real, all fixed — plus the mate now follows its owner.

**`p_clear` was wiping the party frame.** It clears one panel that holds both the party and the mate list, and the party burst was straddling it, so `pinit` and `pst` were cleared on every game start. Both bursts now follow it, in the order the capture shows: `p_clear`, the sc packets, `sc_p_stc`, then `pinit`.

**A newcomer could not see anybody else's mates.** The map-entry loop announced the characters already present but not what was at their heel, so every pet on a populated map was invisible to whoever walked in. Caught by CodeRabbit and entirely right.

**Two rows could claim the same mate slot** — two captures racing, or a hand-edited database — and the second would spawn on top of the first with nothing raised. Rather than a transaction, the reader now decides: the first row keeps the slot and the rest stay in the list. Cheaper, and it repairs a database that is already inconsistent.

**And the mate keeps up.** It moves on the same `CharacterMovedEvent` a step already publishes, taking the first walkable square around the owner, so a mate against a wall tucks in rather than refusing to move and two mates never stack.

denislauri1999:
Two more, both from the outside-diff review, both right.

**Invisibility leaked through the pets.** A hidden owner's mates were still broadcast, and a visible pet carries an `Owner` field naming the character it belongs to — so a GM standing invisible was announced by their own chicken. Spawns and moves now go to the owner alone while they are hidden.

**Both mates were landing on the same square.** Map entry used a fixed `(1,1)` offset, which stacks a pet and a partner and can also be a wall or off the map. The placement the follow already used is now shared (`MatePlacement`): each mate takes its own walkable square, with the owner's own square as the fallback when nothing is free. Two tests — one that the squares are distinct, one that a mate boxed in by walls still lands somewhere walkable.

On the atomic-capture finding I went a different way rather than a transaction: the reader enforces one pet and one partner out at a time. That covers racing captures and also repairs a database that is already inconsistent, which a constraint added now would not.

denislauri1999:
A design question before I go further, because the answer decides the shape of the next PR.

Mate combat (`u_pet`, `u_ps`, `suctl`) means routing a mate through `IBattleService.Hit`, and `IAliveEntity` requires an `Arch.Core.Entity` handle. So a mate that can fight has to be a real ECS entity rather than the plain object it is now.

Which way would you like it?

1. **A `MateComponentBundle`**, alongside monster and npc, created when the mate is summoned and destroyed when it is put away. Mates then get aggro, buffs, cooldowns and death for free, and the plain object stays only as the stored row. More moving parts, and the entity's lifetime has to track the summon.
2. **Keep mates out of the ECS** and give the battle path a narrower interface than `IAliveEntity` for things that can hit and be hit. Smaller now, but it is a second concept of "combatant" in the codebase.

I lean towards 1 — a mate is closer to a monster than to an inventory item, and it is what unblocks the roughly forty-five BCard subtypes that only mean something with a mate in the fight. But it is your architecture, and I would rather ask than send you a large PR built on a guess.

This PR is complete as it stands either way: load, list, spawn, follow, despawn.

denislauri1999:
Thanks for merging NosCoreIO/NosCore.Algorithm#129 — I will delete `MateXpTable` from this PR and take `IMateExperienceService` from the package the moment a version ships. NuGet still shows 2.0.0; I am watching the feed rather than pinging you about it.

denislauri1999:
I answered my own question rather than leave this waiting — **the mate is now an ECS entity and can fight.**

Option 1, the one I said I leaned towards: a `MateComponentBundle` with the same component set a monster has. It turned out to be the cheap answer as well as the right one — `SkillResolver` already branches on `INonPlayableEntity` and reads the creature off `NpcMonster`, so a mate resolves its basic attack exactly as a monster does. No second damage path, and no second notion of "combatant" in the codebase.

- the entity's life is the map's: created on arrival, destroyed on leaving, position written on every step. Leaving it behind strands a mate in a world nobody is in; moving only the packet leaves it visibly in one place and actually in another, which is what a monster picking a target would read.
- `u_pet` checks the mate belongs to the character asking and is actually out — trusting the id would let a client drive somebody else's pet.

**If you would rather mates stayed out of the ECS, say so and it comes straight back out.** The fold of effects is the part that matters and it does not depend on the container. I would rather you reject a branch you can run than answer a question in the abstract.

Still not here: `u_ps` and `suctl` (a partner's skills), loyalty, feeding. Those are their own change.

denislauri1999:
Right again, and it is the mirror of the one from before: I gated the arriving character's own mates on their invisibility but not the loop that tells them about the players already there. A pet's spawn names its owner, so a hidden character was being announced by somebody else's screen. Fixed.

denislauri1999:
Rebased on master. **`MateXpTable` is gone** — `IMateExperienceService` from Algorithm 2.1.0 does it now, and the ten capture observations that pinned the divisors live with the curve as approval tables in that repository.

Thanks for #2287; closing mine was the right call, and bundling the Algorithm bump with it saved a round trip.

src/NosCore.GameObject/Ecs/MapWorld.cs
```diff
@@ -137,6 +137,37 @@ public Entity CreateNpc(
...
+    {
+        var now = SystemClock.Instance.GetCurrentInstant();
+        return World.Create(
+            new EntityIdentityComponent(visualId, VisualType.Npc, mate.CharacterId),
+            new HealthComponent(mate.Hp, mate.MaxHp, true),
+            new ManaComponent(mate.Mp, mate.MaxMp),
+            new PositionComponent(positionX, positionY, direction, mapInstance.MapInstanceId),
+            new VisualComponent(0, 0, 0, 0, false, false, false),
+            new NpcDataComponent(mate.VNum, mate.NpcMonster.Race, mate.Level, 0, mate.NpcMonster.Speed, 10),
+            // A mate never wanders and is never hostile on its own: it goes where its owner
```
erwan-joly: no need for that

src/NosCore.GameObject/Services/MapChangeService/MapChangeService.cs
```diff
@@ -280,6 +324,19 @@ private async Task LeaveMapAsync(ClientSession session)
...
             var channelId = session.Channel!.Id;
             await mapInstance.SendPacketAsync(outPacket, new EveryoneBut(channelId));
+            var leaving = character.Mates.Values.Where(s => s.IsTeamMember).ToList();
+            await mapInstance.SendPacketsAsync(leaving.Select(s => s.GenerateOut()));
+
+            // The entity belongs to the map being left, so it goes with it. A new one is made
+            // on arrival; keeping this one would leave a mate standing in a world nobody is in.
+            foreach (var mate in leaving)
+            {
+                if (mate.Entity is { } handle)
```
erwan-joly: can use where clause instead of imbricated if

src/NosCore.GameObject/Services/MateService/Mate.cs
```diff
@@ -0,0 +1,244 @@
...
+
+        public int MaxMp => NpcMonster.MaxMp;
+
+        /// <summary>
+        /// Written when the mate is loaded rather than computed here: the curve lives in
+        /// NosCore.Algorithm, and a data object has no business resolving a service.
+        /// </summary>
+        public long XpLoad { get; set; }
+
+        public ScpPacket GenerateScp(RegionType language)
```
erwan-joly: the bunch of those packets have nothing to do in the game object

src/NosCore.GameObject/Services/MateService/Mate.cs
```diff
@@ -0,0 +1,244 @@
...
+                Skill3Details = null
+            };
+        }
+
+        /// <summary>
+        /// The spawn packet. Owner and GroupEffect are what tell the client this is somebody's
+        /// mate rather than a map npc:
+        ///     in 2 1506 445562 26 26 2 100 100 0 0 3 626114 1 0 -1 Ratufu^pirate^(Feu) 0 -1 ...
+        /// </summary>
+        public InPacket GenerateIn(RegionType language)
```
erwan-joly: we only need a single way to do all in so that's wrong

erwan-joly (commented):
Likely this can  be split more into smaller PR and also need a good thought on the process as the goal is not just to copy what OpenNos did because that's not the best code possible for lot of those concepts

denislauri1999:
Done, thanks:

- **packets out of the game object** — `GenerateScp/Scn/In/Out/Pst/Cond` are now extension methods in `NosCore.GameObject.Ecs.Extensions.MateExtensions`; `Mate` is data only.
- **`MapWorld.CreateMate` comment** — removed.
- **`MapChangeService` nested if** — now a `Where` clause.

denislauri1999:
Cut the comments right down across all seven open PRs — the remarks blocks are gone and the long explanations are a line each. What I kept is the captured packet lines that pin a field, and the odd note where a reader would otherwise get it wrong; everything that was retelling what the code already says is out.

Also cleaned up a UTF-8 BOM problem I'd introduced: a number of files had picked one up and were showing as whole-line diffs against master. Fixed by matching master per file rather than stripping — the .resx files do carry one upstream, so blanket-stripping was its own kind of noise.

Builds clean, all tests green on each branch.

src/NosCore.GameObject/Ecs/Extensions/MateExtensions.cs
```diff
@@ -0,0 +1,200 @@
...
+                MorphId = mate.Skin != 0 ? mate.Skin : -1,
+                IsSummonable = mate.IsSummonable,
+                SpDetails = null,
+                Skill1Details = null,
+                Skill2Details = null,
+                Skill3Details = null
+            };
+        }
+
+        public static InPacket GenerateIn(this Mate mate, RegionType language)
```
erwan-joly, line 109: just checked the other are also like this so most likely it's for another PR but ideally we only use the visualEntity one so we need to merge them

denislauri1999: Agreed, and I counted them so the follow-up has a shape: there are **six** `GenerateIn` extensions today.

```
INonPlayableEntity visualEntity          VisualEntityExtension.cs   <- the general one
MapItemComponentBundle item              MapItemBundleExtensions.cs
Mate mate, RegionType language           MateExtensions.cs
MonsterComponentBundle monster           MonsterBundleExtensions.cs
NpcComponentBundle npc, short? dialog    NpcBundleExtensions.cs
PlayerComponentBundle player, string prefix   PlayerBundleExtensions.cs
```

The general one already does most of it, and it already special-cases two of the others inline — `visualEntity is INamedEntity` for the name and `visualEntity is NpcComponentBundle` for the dialog. So the merge is mostly a matter of moving the remaining per-type bits behind the same kind of test, or better, behind something the bundles expose themselves.

Three of the five carry an extra argument the general one has no place for yet: the mate's `RegionType` (it feeds `DisplayName`), the npc's `dialog` override, and the player's `prefix`. Those are the part that needs a decision, not just a move.

Happy to take it as its own PR — it touches five files and none of this one's behaviour, so it would only make this one harder to read. Say the word and I'll open it.

src/NosCore.GameObject/Messaging/WolverineDependencyRegistrar.cs
denislauri1999: No, and I should say so plainly: we do not run an official server, so nothing here is checked against one. The base is our choice, not a value read from a data file or a trace.

What can be checked is that it cannot collide. Npc and monster visual ids in this codebase are the database keys — `MapInstance.LoadMonsters` passes `MapMonsterDto.MapMonsterId` and `LoadNpcs` passes `MapNpcDto.MapNpcId` straight into `CreateMonster`/`CreateNpc`. The trace the parser seeds from carries 17 384 `in 3` rows, so those keys stay in the tens of thousands; 2 000 000 is clear of them with three orders of magnitude to spare.

The trace itself gives no support for the number: its `in 2` transport ids run from 1 000 to 7 483 051 with no reserved band anywhere in the range, so an official server does not appear to reserve one either.

The comment that asserted the collision guarantee is gone — the registration is now a bare line like the three above it, and this branch no longer touches the file's existing comments at all.

src/NosCore.GameObject/Services/MapChangeService/MapChangeService.cs
denislauri1999: Gone. `MapChangeService` now carries no added comments at all — `stile-upstream` reports 0 added comment lines on the branch against `origin/master`.

The branch is also rebased onto master (`3cfbde3`) so the history is linear again; the merge commit is gone. Build clean at 0 warnings, 1074 tests green with the CI filter.

src/NosCore.GameObject/Services/MateService/Mate.cs
denislauri1999: Gone, and so is the code it sat on: the four packet generators moved out of `Mate` into `Ecs/Extensions/MateExtensions.cs`, and `Mate` is now a plain data object with five properties.

`stile-upstream` reports 0 added comment lines on the branch.

src/NosCore.GameObject/Services/MateService/Mate.cs
denislauri1999: I took this out first, then measured it, and it turned out the serializer does not do it for these three packets — so it is back, with a test that pins the wire instead of the packet object.

`Serializer.StringSerializer` does hold a `Replace(splitter, "^")`, but the splitter it is handed is `index.SpecialSeparator`, which is null for an ordinary property. `IsNullOrEmpty(splitter)` is then true and the branch returns the value untouched; the space between fields is added later, by `PacketSerializer`, and never enters the substitution. So a plain space-separated field keeps its own spaces.

Serialized with the manual substitution removed, a mate named `Joyeux Mouton`:

```
in    2 333 2000001 0 0 0 0 0 0 0 3 42 1 0 -1 Joyeux Mouton 0 -1 0 0 ...
sc_p  0 333 2000001 0 0 0 ... 1 15 0 Joyeux Mouton 0
sc_n  0 333 2000001 0 0 0 ... 1 15 Joyeux Mouton -1 0 -1 -1 -1 -1
```

Three fields where the client expects one, in all three packets, which shifts everything after the name.

With it back, all three carry `Joyeux^Mouton`. The new test `ASpacedMateNameReachesTheClientAsOneFieldAsync` builds a real `Serializer` over the packet assembly and asserts on the serialized line for each of the three, so this cannot regress silently the way it just did.

Two follow-ups, if you want them — happy either way, and neither belongs in this PR:

- if `Replace` on a property with no `SpecialSeparator` is meant to escape against the field separator rather than against null, that is a fix in `NosCore.Packets` and it would let this line and every other manual substitution in the repo go away;
- there is a second oddity next to it: `isLastIndex` is `indexAttr.Index == maxIndex`, and `maxIndex` is the *outer* packet's highest index, so a sub-packet property whose own index happens to equal it is treated as the packet's last field. `InNonPlayerSubPacket.Name` is index 9 inside an `InPacket` whose last index is also 9.

src/NosCore.GameObject/Services/MateService/MateXpTable.cs
denislauri1999: Done — `MateXpTable.cs` is deleted. `MateService` takes `IMateExperienceService` from `NosCore.Algorithm` and calls `GetPetExperience`/`GetPartnerExperience` on it, so the branch adds no table of its own.

src/NosCore.GameObject/Ecs/MapWorld.cs
denislauri1999: Gone. `CreateMate` now reads like the `CreateNpc` above it, with no added comments — `stile-upstream` reports 0 added comment lines on the branch.

src/NosCore.GameObject/Services/MapChangeService/MapChangeService.cs
denislauri1999: Done:

```csharp
foreach (var mate in leaving.Where(s => s.Entity.HasValue))
{
    mapInstance.EcsWorld.DestroyEntity(mate.Entity!.Value.Handle);
    mate.Entity = null;
}
```

The comment that sat above it is gone too.

src/NosCore.GameObject/Services/MateService/Mate.cs
denislauri1999: Done — all four moved to `Ecs/Extensions/MateExtensions.cs` alongside the other bundle extensions, so `GenerateScp`, `GenerateScn`, `GenerateIn`, `GenerateOut`, `GeneratePst` and `GenerateCond` are extension methods there. `Mate` is left as a data object: five properties and no packet type in sight.

src/NosCore.GameObject/Services/MateService/Mate.cs
denislauri1999: Moved to `Ecs/Extensions/MateExtensions.cs`, so it is now one more `GenerateIn` extension alongside the five that already exist rather than a method on the game object.

That leaves the real point, which is the one you raised again later on `MateExtensions.cs:109`: there are six of them and they should be one. I counted and answered there — three of the five carry an extra argument the general `INonPlayableEntity` one has no place for yet (the mate's `RegionType`, the npc's dialog override, the player's prefix), and that is the part needing a decision rather than a move. Happy to take it as its own PR; it touches five files and none of this one's behaviour.

PR #2283, opened 2026-08-23
feat(family): a character's family is loaded, named, and described

denislauri1999:
The `Family`, `FamilyCharacter` and `FamilyLog` tables have been in the schema since the first migration and nothing has ever read them. A character who belonged to a family looked — to everyone including themselves — like a character who did not.

This adds the reading half: the membership is loaded at character select, the family tag goes into `c_info`, and the `ginfo` window is sent at login. 9 tests, all green, no new warnings.

## Where the shapes come from

A packet capture, field for field.

* **`ginfo` is confirmed all seventeen fields deep.** Worth stating because seventeen numbers in a row is exactly where an off-by-one hides, and this one would hand out the wrong permissions without throwing anything:

  ```
  ginfo -Nemesis- Yzigor 0 7 130000 640000 68 70 3 1 1 1 1 2 1 2 coin^afk^go^rush
  ```

* **The family tag is `Name(Rank)`** — the brackets some families have are part of the *name*: `[NDM](Gardien)`, `KillaBeez(Gardien)`, `Survival(Membre)`.
* **`c_info`** carries the family id and that tag, and `-1` with a bare `-` when there is none.

## Two things found along the way

* **`c_info` field 15 is not the family level**, whatever the packet type calls it. PIKAZ in the capture belongs to a family `gidx` says is level 10, and their `c_info` carries `0` in that slot. Left at zero rather than filled with something the capture contradicts.
* **The inherited family experience table disagrees with the capture.** It says a level-7 family needs 1 900 000 to advance; the captured `ginfo` says 640 000 — nearly three times less. One observation cannot rebuild nineteen rows, so the inherited numbers stay and the disagreement is written into the table's own comment. It only draws the bar; nothing levels a family up yet.

## What is deliberately missing

**`gidx`**, the tag over a character's head — blocked rather than skipped.

The capture writes its third field as a single family id (670 lines, not one containing a dot) and `-1` when there is no family. `GidxFamilySubPacket` models it as a `serverId.familyId` compound, which cannot produce either shape. Leaving the sub-packet null makes the serializer drop the separating space and emit `gidx 1 741328-1 - 0`, which the client cannot split into fields.

The same missing space happens on `sc_n`'s empty equipment slots, so this looks like **one defect in `NosCore.Packets` showing twice**. Happy to send a change there for both if you want it.

## A note on placement

The family lives in `PlayerSocialComponent` rather than a component of its own: it is social state, and Arch's `World.Create` is out of generic overloads at the player bundle's current size — a further component would have to be added separately and would be easy to forget on the map-change path.

## Summary by CodeRabbit

- **New Features**
  - Added family authority roles: Head, Assistant, Manager, and Member.
  - Family information now includes member roles, family details, and the head’s name.
  - Family tags and information windows display localized authority ranks.
  - Family details load when selecting a character and at game startup.
  - Improved character visibility during map transitions.

- **Bug Fixes**
  - Players without a family or with missing family data are handled gracefully.


src/NosCore.Data/Resource/LocalizedResources.it.resx
```diff
@@ -574,4 +574,16 @@
   <data name="UNHANDLED_UPGRADE_TYPE" xml:space="preserve">
     <value>Tipo di potenziamento non gestito: {0}</value>
   </data>
+  <data name="FAMILY_AUTHORITY_HEAD" xml:space="preserve">
```
erwan-joly, line 604: all those are not translated we shouldnt add a value and not  translate them

src/NosCore.GameObject/Services/FamilyService/FamilyExperienceTable.cs
```diff
@@ -0,0 +1,63 @@
...
+// |_|\__|\__/ |___/ \__/\__/|_|_\___|
+//
+
+namespace NosCore.GameObject.Services.FamilyService
+{
+    /// <summary>
+    /// How much experience a family needs to leave its current level.
+    /// </summary>
+    /// <remarks>
+    /// INHERITED, AND THE ONE THING WE CAN CHECK SAYS IT IS WRONG. The table below is the one the
```
erwan-joly: again lot of those comments are useless

src/NosCore.GameObject/Services/FamilyService/FamilyExperienceTable.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </remarks>
+    public static class FamilyExperienceTable
+    {
+        /// <summary>
+        /// Experience to leave <paramref name="familyLevel" />, or a number the bar can never
+        /// fill once past the last described level.
+        /// </summary>
+        public static uint RequiredExperience(byte familyLevel)
+        {
+            return familyLevel switch
```
erwan-joly: that shouldn't be in noscore but in the library for stats

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            var rank = authority switch
+            {
+                FamilyAuthority.Head => LanguageKey.FAMILY_AUTHORITY_HEAD,
+                FamilyAuthority.Assistant => LanguageKey.FAMILY_AUTHORITY_ASSISTANT,
+                FamilyAuthority.Manager => LanguageKey.FAMILY_AUTHORITY_MANAGER,
+                _ => LanguageKey.FAMILY_AUTHORITY_MEMBER
+            };
+
+            return $"{Name}({localizer[rank, language]})".Replace(' ', '^');
```
erwan-joly: that should be handled by the packet serialization

denislauri1999:
Comments cut — 145 lines gone.

**Translations:** the four rank labels are now translated in all eight non-English resources instead of sitting there in English. The French is not a guess: the capture writes the tag as `[NDM](Gardien)`, so *Gardien* is what Manager comes out as on a real server, and the other seven follow that reading.

**"that shouldn't be in noscore but in the library for stats":** agreed — NosCoreIO/NosCore.Algorithm#129 adds `FamilyExperienceService` (with `MateExperienceService` for the same complaint on #2281). `FamilyExperienceTable` stays here only to keep this branch building and goes as soon as that package ships; happy to hold this PR until then instead.

Worth flagging from that work: the published family table says a level 7 family needs 1 900 000 to advance, and the captured `ginfo` for a level 7 family says 640 000. One observation cannot rebuild eighteen rows so I changed nothing, but that row is wrong.

**"that should be handled by the packet serialization":** yes. Same conclusion as on #2281 — the space-to-caret substitution belongs in the serializer, and while looking I found a second thing there: a null sub-packet loses its separating space, which is why `gidx` cannot currently be sent at all. The capture writes its family field as a single id (670 lines, no dots) and `-1` when there is none, while `GidxFamilySubPacket` models it as `serverId.familyId`, which produces neither. Both go into one NosCore.Packets PR and the manual `Replace` comes out of here once it lands.

denislauri1999:
Same as on #2281: with NosCoreIO/NosCore.Algorithm#129 merged, `FamilyExperienceTable` comes out of this PR in favour of `IFamilyExperienceService` as soon as a package version is published.

src/NosCore.GameObject/Ecs/Extensions/PlayerBundleExtensions.cs
```diff
@@ -434,8 +433,11 @@ public static CInfoPacket GenerateCInfo(this PlayerComponentBundle player, IGame
                 ? $"[{localizer[LanguageKey.SUPPORT, player.Account.Language]}]" + player.Name : player.Name,
             Unknown1 = null,
             GroupId = -1,
-            FamilyId = -1,
-            FamilyName = null,
+            FamilyId = (int)(player.FamilyCharacter?.FamilyId ?? -1),
+            FamilyName = player.FamilyCharacter == null
```
erwan-joly: this should be handled by the  serializer already

src/NosCore.GameObject/Ecs/Extensions/PlayerBundleExtensions.cs
```diff
@@ -434,8 +433,11 @@ public static CInfoPacket GenerateCInfo(this PlayerComponentBundle player, IGame
                 ? $"[{localizer[LanguageKey.SUPPORT, player.Account.Language]}]" + player.Name : player.Name,
             Unknown1 = null,
             GroupId = -1,
-            FamilyId = -1,
-            FamilyName = null,
+            FamilyId = (int)(player.FamilyCharacter?.FamilyId ?? -1),
```
erwan-joly: likely this need to be nullable in the packet so we don't have to return -1 which is default for null

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,66 @@
...
+                FamilyManagerCanNotice = ManagerCanNotice,
+                FamilyManagerCanShout = ManagerCanShout,
+                FamilyManagerCanGetHistory = ManagerCanGetHistory,
+                FamilyManagerAuthorityType = ManagerAuthorityType,
+                FamilyMemberCanGetHistory = MemberCanGetHistory,
+                FamilyMemberAuthorityType = MemberAuthorityType,
+                // The last field of a packet is left alone by the serializer, on the grounds
+                // that it may legitimately hold spaces. A family message may not: the capture
+                // shows it caret-separated.
+                FamilyMessage = (FamilyMessage ?? string.Empty).Replace(' ', '^')
```
erwan-joly: again shouldnt replace manually by carret I believe we can do that in the serialization already

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            var rank = authority switch
+            {
+                FamilyAuthority.Head => LanguageKey.FAMILY_AUTHORITY_HEAD,
+                FamilyAuthority.Assistant => LanguageKey.FAMILY_AUTHORITY_ASSISTANT,
+                FamilyAuthority.Manager => LanguageKey.FAMILY_AUTHORITY_MANAGER,
+                _ => LanguageKey.FAMILY_AUTHORITY_MEMBER
+            };
+
+            return $"{Name}({localizer[rank, language]})";
```
erwan-joly: this should likely get the language from the user that receive the packet

src/NosCore.GameObject/Services/FamilyService/FamilyCharacter.cs
```diff
@@ -0,0 +1,17 @@
...
+// |_|\__|\__/ |___/ \__/\__/|_|_\___|
+//
+
+using NosCore.Data.Dto;
+
+namespace NosCore.GameObject.Services.FamilyService
+{
+    public class FamilyCharacter : FamilyCharacterDto
+    {
+        public string CharacterName { get; set; } = string.Empty;
```
erwan-joly: what is this for ? we likely don't need it

src/NosCore.GameObject/Services/FamilyService/FamilyExperienceTable.cs
```diff
@@ -0,0 +1,38 @@
...
+
+namespace NosCore.GameObject.Services.FamilyService
+{
+    public static class FamilyExperienceTable
+    {
+        public static uint RequiredExperience(byte familyLevel)
+        {
+            return familyLevel switch
+            {
+                1 => 100_000,
```
erwan-joly: those stats were wrong also fixed to lvl 30

src/NosCore.PacketHandlers/Game/GameStartPacketHandler.cs
```diff
@@ -154,6 +154,13 @@ await session.SendPacketAsync(new TwkPacket(session.Account.Name, session.Charac
             //            Session.SendPackets(Session.Character.GenerateScN());
             //            Session.Character.GenerateStartupInventory();
 
+            var membership = session.Character.FamilyCharacter;
+            if (membership != null)
+            {
+                await session.SendPacketAsync(membership.Family.GenerateGInfo(membership.Authority,
+                    FamilyExperienceTable.RequiredExperience(membership.Family.FamilyLevel))).ConfigureAwait(false);
```
erwan-joly: likely this should be inside the GenerateGInfo as a experienceService

erwan-joly: also don't need the configure await

src/NosCore.PacketHandlers/Game/GameStartPacketHandler.cs
```diff
@@ -154,6 +154,13 @@ await session.SendPacketAsync(new TwkPacket(session.Account.Name, session.Charac
             //            Session.SendPackets(Session.Character.GenerateScN());
             //            Session.Character.GenerateStartupInventory();
 
+            var membership = session.Character.FamilyCharacter;
+            if (membership != null)
```
erwan-joly: familyMembership would be a better name

denislauri1999:
Rebased on master and took both bumps.

**`gidx` is in.** Flattening the family id in Packets 21.0.0 was the blocker, and both cases now reproduce the capture exactly:

```
gidx 1 521919 5083 [NDM](Gardien) 3
gidx 1 741328 -1 - 0
```

It goes out on map entry — the arriving character's tag to everybody, everybody else's to the newcomer. A character with no family is told so rather than left alone: the client keeps whatever tag it was last given, so silence leaves a stale one over their head.

**`FamilyExperienceTable` is gone**, replaced by `IFamilyExperienceService` from Algorithm 2.1.0. The disagreement with the capture — 1 900 000 against the captured 640 000 at level 7 — travelled with the curve and is written on it there.

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// What the client is told when a character has no family at all.
+        /// </summary>
+        public static GidxPacket GenerateEmptyGidx(long visualId)
+        {
+            return new GidxPacket
+            {
+                VisualType = VisualType.Player,
+                VisualId = visualId,
+                FamilyId = null,
+                FamilyName = "-",
```
erwan-joly: that should be null (handled  by serializer)

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,102 @@
...
+                FamilyId = FamilyId,
+                FamilyName = GenerateFamilyTag(authority, localizer, language),
+                FamilyLevel = FamilyLevel
+            };
+        }
+
+        /// <summary>
+        /// What the client is told when a character has no family at all.
+        /// </summary>
+        public static GidxPacket GenerateEmptyGidx(long visualId)
```
erwan-joly: this shouldn't be here (need to follow our normal ecs conventions)

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,102 @@
...
+
+        /// <summary>
+        /// The family tag drawn over a character's head.
+        /// </summary>
+        /// <remarks>
+        /// A capture writes the family field as a single id in all 670 of its gidx lines, and
+        /// -1 where there is no family — which is what the packet now models, since 21.0.0.
+        ///     gidx 1 521919 5083 [NDM](Gardien) 3
+        /// </remarks>
+        public GidxPacket GenerateGidx(long visualId, FamilyAuthority authority,
```
erwan-joly: those are for the ecs

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,102 @@
...
+namespace NosCore.GameObject.Services.FamilyService
+{
+    public class Family : FamilyDto
+    {
+        public IReadOnlyList<FamilyCharacter> Members { get; set; } = [];
+
+        public FamilyCharacter? Head =>
+            Members.FirstOrDefault(s => s.Authority == FamilyAuthority.Head);
+
+        public GInfoPacket GenerateGInfo(FamilyAuthority readerAuthority, uint experienceToNextLevel)
```
erwan-joly: this is for the ecs shouldnt be inside a game object

src/NosCore.GameObject/Services/FamilyService/FamilyService.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace NosCore.GameObject.Services.FamilyService
+{
+    public class FamilyService(IDao<FamilyDto, long> familyDao,
+        IDao<FamilyCharacterDto, long> familyCharacterDao,
+        IDao<CharacterDto, long> characterDao) : IFamilyService
+    {
+        public async Task<FamilyCharacter?> GetMembershipAsync(long characterId)
```
erwan-joly: shouldn't we just get the Family not sure we need this concept of FamilyCharacter at all tbh

denislauri1999:
All ten addressed in one pass.

**The ECS conventions** — you were right that the packets had no business on the game object. They are now an extension on the player bundle, next to `GenerateCInfo`; `Family` and `FamilyCharacter` are data again.

**The language one was a real bug, not just a smell.** The tag was built from the *owner's* account language, so two players standing next to each other saw the same family in whichever language its owner happens to play in. It reads the receiver's language now, with a test — the words live in the resources, so what the test guards is that the tag is built per reader at all.

The rest:

* the empty tag is `null` rather than `"-"`, and the caret substitution on the family message is gone — both duplicated a rule that lives in the serializer;
* `FamilyCharacter.CharacterName` removed. You were right that we do not need it: only the head's name is ever printed, so only the head's name is fetched and it sits on the family;
* the experience service goes into `GenerateGInfo` rather than being resolved at the call site;
* `familyMembership`, and no `ConfigureAwait`.

And thank you for fixing the family curve — that settles the disagreement I flagged (640 000 at level 7 against the table's 1 900 000) and takes it to level 30, which the old one did not reach.

erwan-joly:
PR can likely be broken down more

denislauri1999:
You're right, `FamilyCharacter` wasn't earning its place — dropped it.

`PlayerSocialComponent` now holds the `Family` directly, `Family.Members` is a plain `IReadOnlyList<FamilyCharacterDto>`, and the reader's own rank comes from `Family.AuthorityOf(characterId)`. That also removes the cycle where each membership pointed back at its family. `IFamilyService.GetMembershipAsync` is now `GetFamilyAsync`, and `GameStartPacketHandler` just null-checks the packet `GenerateGInfo` returns instead of checking the membership first.

Kept a test on `AuthorityOf` for the case that matters: the window hands out permissions from that value, so reading somebody else's row would grant them without throwing.

denislauri1999:
Good catch from the bot, and it was mine: I fixed the per-reader language inside `GenerateGidx` and left the call site that broadcasts it.

`MapChangeService` now sends one `gidx` per watcher, each in that watcher's `AccountLanguage`. The `GenerateIn` above it stays as it was — its prefix is the moderator tag, which belongs to the character rather than the reader. The other `GenerateGidx` in the file (line 228) was already right: that one sends *to* the arriving player, so their own language is the correct one.

Also added while master is now on Packets 21.0.0: a test that asserts the serialized line rather than the packet object, since the line is what the client reads —

    gidx 1 741328 -1 - 0

A null id has to reach the client as `-1` and a null name as `-`, and asserting on the object would have passed whatever the serializer then did with it.

src/NosCore.Data/Enumerations/I18N/LanguageKey.cs
```diff
@@ -189,6 +189,10 @@ public enum LanguageKey
         OBJECT_IN_BAZAAR,
         ITEM_GIFTED,
         NOT_IN_FRIENDLIST,
-        SPEAKER
+        SPEAKER,
+        FAMILY_AUTHORITY_HEAD,
```
erwan-joly, line 202: interesting so those don't have a internationalized version on official?

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Linq;
+
+namespace NosCore.GameObject.Services.FamilyService
+{
+    public class Family : FamilyDto
+    {
+        public IReadOnlyList<FamilyCharacterDto> Members { get; set; } = [];
+
+        /// <summary>The head's name, kept here because the head is usually offline.</summary>
+        public string HeadCharacterName { get; set; } = string.Empty;
```
erwan-joly, line 19: we need a family member name lookup as it's not the only name we need. So likely we should just rely on Member

src/NosCore.GameObject/Services/FamilyService/Family.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        /// <summary>Not in the member list means Member: the packet cannot say "none".</summary>
+        public FamilyAuthority AuthorityOf(long characterId) =>
```
erwan-joly, line 22: why would somebody not be in member list?

src/NosCore.Data/Resource/LocalizedResources.it.resx
denislauri1999: Translated, and then corrected — the first pass invented them, which is worse than leaving them in English because nobody would have looked again.

The client already has these four. `conststring` holds them at 10915-10918, parentheses included, in every language it ships:

```
UK  (Familyhead)  (Familydeputy)  (Familykeeper)  (Member)
FR  (Tête)        (Assistant)     (Gardien)       (Membre)
DE  (Oberhaupt)   (Vertreter)     (Verwalter)     (Mitglied)
IT  (Capo)        (Vicecapo)      (Guardiano)     (Membro)
ES  (Cabecilla)   (Representante) (Administrador) (Miembro)
CZ  (Hlava)       (Zástupce)      (Správce)       (Člen)
PL  (Szef)        (Reprezentant)  (Zarządca)      (Członek)
TR  (Ailereisi)   (Temsilci)      (Yönetici)      (Üye)
```

and a captured line reads `gidx 1 626114 5052 -Nemesis-(Membre) 8`, which is the French row exactly.

So the values are the client's wording now, not a translation of ours. **Eighteen of the thirty-two were wrong** — the first French one said `Chef de famille` where the client says `Tête`, the English one said `Head` against `Familyhead`. Russian keeps ours: the client ships no Russian file, so there is nothing to read.

src/NosCore.Data/Enumerations/I18N/LanguageKey.cs
denislauri1999: They do, and I should have checked before adding four keys. `conststring` carries the four ranks at 10915-10918, with the parentheses already in the string, in every language the client ships:

```
UK  (Familyhead)  (Familydeputy)  (Familykeeper)  (Member)
FR  (Tête)        (Assistant)     (Gardien)       (Membre)
DE  (Oberhaupt)   (Vertreter)     (Verwalter)     (Mitglied)
IT  (Capo)        (Vicecapo)      (Guardiano)     (Membro)
ES  (Cabecilla)   (Representante) (Administrador) (Miembro)
CZ  (Hlava)       (Zástupce)      (Správce)       (Člen)
PL  (Szef)        (Reprezentant)  (Zarządca)      (Członek)
TR  (Ailereisi)   (Temsilci)      (Yönetici)      (Üye)
```

A captured line, `gidx 1 626114 5052 -Nemesis-(Membre) 8`, matches the French row — so the server sends the rank as text, and the client is not resolving a number here the way it does for `Game18NConstString`.

I have put the client's own wording into the resources rather than our translations, which fixed eighteen of the thirty-two values. What I have **not** done is move the source: they are still four `LanguageKey`s rather than rows imported from `conststring`. `ConstStringParser` reads only the UK file today, on the stated grounds that the other languages wrap the same numbers in translated prose — which is true for the ladders but not for these, where the text *is* the payload and differs per language by design. Turning that into a per-language import is a change to the parser's shape, so I would rather do it as its own PR if you want it. Say the word.

src/NosCore.GameObject/Services/FamilyService/FamilyExperienceTable.cs
denislauri1999: Gone with the file — `FamilyExperienceTable.cs` is deleted and the curve comes from `IFamilyExperienceService` in the stats library. `stile-upstream` reports 0 added comment lines on the branch against `origin/master`.

src/NosCore.GameObject/Services/FamilyService/FamilyExperienceTable.cs
denislauri1999: Done — the table is deleted. `GenerateGInfo` takes `IFamilyExperienceService` and calls `GetFamilyExperience(family.FamilyLevel)`, so the branch adds no curve of its own.

src/NosCore.GameObject/Services/FamilyService/FamilyExperienceTable.cs
denislauri1999: Taken — the branch no longer carries a table at all. `FamilyExperienceTable.cs` is deleted and the values come from `IFamilyExperienceService` in the stats library, so whatever is right lives in one place and this PR is not a second opinion on it.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: I removed it, then measured, and it has to come back — the serializer does not do this for an ordinary property.

`Serializer.StringSerializer` holds a `Replace(splitter, "^")`, but the splitter it is handed is `index.SpecialSeparator`, which is null for a plain property. `IsNullOrEmpty(splitter)` is then true and the value comes back untouched; the space between fields is added afterwards by `PacketSerializer` and never enters the substitution.

Measured with the substitution removed:

```
ginfo -Nemesis- Yzigor 0 7 130000 512000 1 70 0 1 1 1 1 0 1 0 coin afk go rush
```

against the captured line, which is

```
ginfo -Nemesis- Yzigor 0 7 130000 640000 68 70 3 1 1 1 1 2 1 2 coin^afk^go^rush^larene^30^min^=p
```

Four fields where the client expects one. The tag is escaped for the same reason; a family name is free text out of the database.

The comment that used to sit above this is gone, and the new test asserts on the **serialized line** rather than the packet object — asserting on the object is what let this through in the first place.

If that `Replace` in the serializer is meant to guard the field separator rather than a special one, fixing it in `NosCore.Packets` would let this line and every other manual substitution in the tree go away. Happy to open that if you want it; it is not something I want to change under this PR.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: I removed it, then measured, and it has to come back — the serializer does not do this for an ordinary property.

`Serializer.StringSerializer` holds a `Replace(splitter, "^")`, but the splitter it is handed is `index.SpecialSeparator`, which is null for a plain property. `IsNullOrEmpty(splitter)` is then true and the value comes back untouched; the space between fields is added afterwards by `PacketSerializer` and never enters the substitution.

Measured with the substitution removed:

```
ginfo -Nemesis- Yzigor 0 7 130000 512000 1 70 0 1 1 1 1 0 1 0 coin afk go rush
```

against the captured line, which is

```
ginfo -Nemesis- Yzigor 0 7 130000 640000 68 70 3 1 1 1 1 2 1 2 coin^afk^go^rush^larene^30^min^=p
```

Four fields where the client expects one. The tag is escaped for the same reason; a family name is free text out of the database.

The comment that used to sit above this is gone, and the new test asserts on the **serialized line** rather than the packet object — asserting on the object is what let this through in the first place.

If that `Replace` in the serializer is meant to guard the field separator rather than a special one, fixing it in `NosCore.Packets` would let this line and every other manual substitution in the tree go away. Happy to open that if you want it; it is not something I want to change under this PR.

src/NosCore.GameObject/Ecs/Extensions/PlayerBundleExtensions.cs
denislauri1999: Done for the name — the branch no longer spells out the empty case. `GenerateCInfo` reads

```csharp
FamilyName = player.GenerateGidx(localizer, player.AccountLanguage).FamilyName,
```

and that is `null` when there is no family, which the serializer writes as `-`. A test pins the wire form: `gidx 1 <id> -1 - 0`, matching the captured `gidx 1 792103 -1 - 0`.

The `FamilyId` half of the same line is answered in your other comment on it.

src/NosCore.GameObject/Ecs/Extensions/PlayerBundleExtensions.cs
denislauri1999: You are right about the shape, and I have left it alone on purpose — it is a change in `NosCore.Packets`, not here.

`CInfoPacket.FamilyId` is `int`, while `GroupId` on the line above it is already `short?`, so the two neighbouring sentinels are typed differently for no reason. Making it `int?` would let this pass `null`: `Serializer.NullableSerializer` writes `-1` for a null non-optional value, so the bytes on the wire would be identical to what the line produces today.

That is the chain — branch, PR, CI, merge, tag, nuget, bump — so I would rather not start it inside this one. Say the word and I will open it against `NosCore.Packets`, and bump here once it is indexed.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: Done — it is `null` now, not `"-"`, and `GenerateEmptyGidx` is gone as a separate method: `GenerateGidx` returns the empty shape itself when `player.Family` is null.

`TheTagGoesOutOnTheWireExactlyAsTheCaptureHasIt` pins the result on the wire rather than on the object, and asserts the line ends in ` -1 - 0` — the serializer's own rendering of a null id and a null string, and what the captured `gidx 1 792103 -1 - 0` shows.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: Done — the tag is built in the reader's language, not the subject's:

```csharp
public static GidxPacket GenerateGidx(this PlayerComponentBundle player,
    IGameLanguageLocalizer localizer, RegionType viewerLanguage)
```

and every broadcast passes the receiving session's language. `TheRankIsReadInTheLanguageOfWhoeverIsLooking` asserts the EN and FR tags for the same character differ.

src/NosCore.GameObject/Services/FamilyService/FamilyCharacter.cs
denislauri1999: Deleted — `FamilyCharacter.cs` is gone, along with the extra name it carried. Answered more fully on your `FamilyService` comment.

src/NosCore.PacketHandlers/Game/GameStartPacketHandler.cs
denislauri1999: Both done. `GenerateGInfo` is an extension on the bundle taking `IFamilyExperienceService`, so the handler no longer computes anything:

```csharp
await session.SendPacketAsync(session.Character.GenerateGInfo(familyExperienceService));
```

and it returns `null` when the character has no family, so the `if` is gone too. The `ConfigureAwait` is gone — and so are the other three the branch had added, following the rule you added in #2335.

src/NosCore.PacketHandlers/Game/GameStartPacketHandler.cs
denislauri1999: The local is gone rather than renamed — `FamilyCharacter` was dropped, so the handler reads `session.Character.Family` and there is no membership variable left to name.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: Moved — every packet generator is now in `Ecs/Extensions/FamilyExtensions.cs` as an extension on `PlayerComponentBundle`, alongside the other bundle extensions. `Family` is a data object again: a member list, the head's name, and one lookup.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: Moved — every packet generator is now in `Ecs/Extensions/FamilyExtensions.cs` as an extension on `PlayerComponentBundle`, alongside the other bundle extensions. `Family` is a data object again: a member list, the head's name, and one lookup.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: Moved — every packet generator is now in `Ecs/Extensions/FamilyExtensions.cs` as an extension on `PlayerComponentBundle`, alongside the other bundle extensions. `Family` is a data object again: a member list, the head's name, and one lookup.

src/NosCore.GameObject/Services/FamilyService/FamilyService.cs
denislauri1999: Done — `FamilyCharacter` is deleted and the service returns the `Family`:

```csharp
Task<Family?> GetFamilyAsync(long characterId);
```

`character.Family` is what the rest of the code holds now, and the membership row only survives as an entry in `Family.Members`, which the ginfo member count and the authority lookup both need anyway. That is one type fewer and one indirection fewer at every call site.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: Agreed, and it is worth settling before anything else reads a member.

`FamilyCharacterDto` carries `CharacterId` and no name — the name lives on `Character`, one table across. So today `GetFamilyAsync` loads the membership rows and then does one extra `characterDao` lookup for the head, because ginfo needs exactly one name and no other packet on this branch needs any.

The moment the member list window arrives that becomes N lookups, which is the wrong shape. Two ways out, and I would rather you pick than guess:

1. project members to a small record carrying `CharacterId`, `Authority`, `Rank`, `Experience` and `Name`, filled by one `characterDao.Where(s => ids.Contains(s.CharacterId))` for the whole family. `HeadCharacterName` then disappears and ginfo reads the head off `Members`.
2. leave `Members` as the raw rows and let whoever draws the list resolve names, which keeps this PR unchanged but repeats the join at every call site.

I lean on (1) and it is a small change — the service already has the id list in hand. Happy to put it in this PR or keep it for the one that adds the window; say which.

src/NosCore.GameObject/Services/FamilyService/Family.cs
denislauri1999: They would not — and you are right that the fallback is hiding something.

`GetFamilyAsync` finds the family *through* this character's own membership row, so by the time `Family` exists the character is in `Members` by construction. The `?? FamilyAuthority.Member` is there only because `AuthorityOf` takes an arbitrary id and the expression has to be total; the branch is unreachable in every call the codebase makes.

The redundancy underneath it is real: the service reads the membership row, drops it, and `AuthorityOf` then finds it again in the list. It should either return the authority it already has, or `AuthorityOf` should be gone and the caller should read it off the row.

That is tangled with the other comment about `Members` carrying names, so I have left both alone rather than half-doing them — tell me which shape you want for `Members` and I will do the pair in one go.

PR #2288, opened 2026-08-23
Add the BCard subtypes the files declare and the enums stop before

denislauri1999:
BCard.dat declares ten subtypes for type 104; `AdditionalTypes.Type104` had six, and the pair it marked `//SubType 4 didn't exist.` is declared like every other one — the same shape as the note that used to sit on type 94's subtype 41.

    41  "Increases the attack power of your NosMate by %s%%."

Two of the additions are **not** negations, so they are named for what they say rather than with a `Negated` suffix:

| subtype | what the file says |
|---|---|
| `11` | reflect a share of the damage, capped at 50% of the carrier's max HP |
| `12` | the same sentence **without** the cap — not a negation of 11 |
| `51` | "All opponents within %s space(s) suffer %s every 1.5 seconds." |
| `52` | "**Allies** within %s space(s) suffer [%s] every 1.5 seconds." |

A handler treating 52 as 51 negated would hang a hostile aura on its own party.

I left the names of `12` and `51` alone so nothing downstream breaks; what they actually mean is now written on the enum.

Enum only — no behaviour change. Zero warnings, all tests pass.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the classification of X2 slot effects.
  * Expanded descriptions for additional combat-effect subtypes.
  * Corrected the label for the negated companion-attack increase effect to improve clarity and accuracy.

### DIFF
--- a/src/NosCore.Data/Enumerations/Buff/AdditionalTypes.cs
+++ b/src/NosCore.Data/Enumerations/Buff/AdditionalTypes.cs
@@ -113,12 +113,7 @@ namespace NosCore.Data.Enumerations.Buff
             ChanceMagicalDecreased = 42,
 
             /// <summary>
-            /// "Final damage from incoming critical hits is reduced by %s%% per critical hit
-            /// (max. %s hits)." Not a negation of anything above it: the first four pairs are the
-            /// parry, which is a chance, while this is priced per critical in a RUN.
-            ///
-            /// 52 carries that same sentence word for word, so the suffix marks the X2 slot and
-            /// not an inversion. Nothing in the game data declares it today.
+            /// Per-critical reduction over a run of hits, not a chance like 11-42.
             /// </summary>
             CriticalDamageReducedPerHit = 51,
             CriticalDamageReducedPerHitNegated = 52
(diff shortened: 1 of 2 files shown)


denislauri1999:
Added a second one of the same kind while working through the coverage, so it does not need a PR of its own:

**`AdditionalTypes.Block` (type 11) stops at 42, and the file declares a fifth pair.**

    11/51-52  "Final damage from incoming critical hits is reduced by %s%% per critical hit (max. %s hits)."

It is **not** a negation of anything above it. Type 11's first four pairs are the parry — a chance that the damage changes — while this is priced *per critical in a run*, which is a different shape and the reason it needs its own name instead of being folded into `ChanceAllDecreased`.

Retitled the PR accordingly. Still enum-only, no behaviour change, zero warnings, all tests pass.

denislauri1999:
Checked both findings against BCard.dat, resolved through the client's language files.

**The count** — right, five members were added to `Type104` (22, 32, 41, 42, 52), not four. Fixed. A UTF-8 BOM had also crept into the file and was showing up as a whole-line diff against master; that's gone too.

**The names** — the underlying observation is correct. The client repeats the X1 sentence word for word for all three:

```
104/21  All opponents within %s space(s) take %s damage every 1.5 seconds.
104/22  All opponents within %s space(s) take %s damage every 1.5 seconds.

104/41  Increases the attack power of your NosMate by %s%%.
104/42  Increases the attack power of your NosMate by %s%%.

 11/51  Final damage from incoming critical hits is reduced by %s%% per critical hit (max. %s hits).
 11/52  Final damage from incoming critical hits is reduced by %s%% per critical hit (max. %s hits).
```

So nothing in the files says these invert.

`MateAttackDecreased` asserted a decrease outright, which the file does not support — renamed to `MateAttackIncreasedNegated`, the base+`Negated` form used elsewhere in this file.

I've kept `Negated` on the other two rather than inventing a neutral suffix, because it already *is* the neutral suffix here: `Type107.MagicArmourFlatNegated` (11/12) is exactly this case — identical text on both slots — and there are 154 `*Negated` members in the file. A new suffix for three of them would make these additions the odd ones out. What was missing was saying so, so the remarks now record that 22/42/52 carry the X1 sentence unchanged, that the suffix marks the slot rather than the effect, and that no skill, card or item declares any of the three today.

Where the two sentences genuinely differ, the names say what the difference is rather than using the suffix — `SummonTwoMonstersOnDef` (32) and `AreaBuffOnAlliesEachSecond` (52), the latter being the one where reading 52 as the negation of 51 would put a hostile aura on your own party.

denislauri1999:
Added type 19 (`Move`), which is the same problem this PR is about.

The files say:

```
11/12  Movement impossible                             (same text in both slots)
21/22  Movement speed is increased/decreased by %s%%
31/32  Your movement speed is increased/decreased by %s while you are hidden
41/42  Movement speed is increased/decreased by %s
51/52  Your speed is temporarily maximised             (same text in both slots)
```

`SetMovement` / `SetMovementNegated` for 31-32 says nothing about a percentage, a flat value, or being hidden — and 31-32 is the one pair that is conditional. Renamed to `SpeedWhileHiddenIncreased` / `SpeedWhileHiddenDecreased`; nothing referenced the old names.

I left 21-42 named as they are and documented the trap instead: 21-22 is a **percentage**, 41-42 is **flat**, and `MoveSpeedIncreased` / `MovementSpeedIncreased` differ by one syllable.

Why it is worth the churn — the placeholder in `SpeedCalculationService` reads:

```csharp
var bonusSpeed = 0; /*(byte)GetBuff(CardType.Move, (byte)AdditionalTypes.Move.SetMovementNegated)[0];*/
```

Subtype 32 is *"movement speed is **decreased** by %s while you are hidden"*. Uncommented as it stood, that line would have applied a conditional decrease as a general speed bonus. I replaced it with a note saying which pair is which — actually reading them is a separate change.

PR #2289, opened 2026-08-23
Skills with a CELL pattern hit their pattern, not a single target

denislauri1999:
Sixty-eight skills in `Skill.dat` carry a `CELL` section: an explicit list of the cells they hit, relative to the caster and drawn facing north. **Sixty-seven of them are declared as area (HitType 3) with an area radius of ZERO** — because the area *is* the pattern, not the radius.

`TargetResolver` only read the radius, so all sixty-seven hit a single target. No exception, no log, just less damage than the file says:

| skill | what the file draws | what it did |
|---|---|---|
| 244 Piercing Shot | a row of eight cells in front — an arrow that pierces | hit one |
| 367 Fire Breath | a rectangle of thirty | hit one |

### The rotation

The pattern is authored facing north and cast towards the target, so it is rotated by the angle between the two rather than snapped to eight directions. On the four cardinals the result is exact (sine and cosine are 0 or ±1); on the diagonals it is the nearest approximation.

**An approximation worth stating**, since it is visible in play: rotating by an arbitrary angle and rounding can turn a row one cell wide into a staircase with the odd gap at its side, and somebody half a cell off the line can slip through. On the cardinals — the overwhelming majority of casts, since the caster lines up with the target — it does not arise.

### The table

Generated from the client’s `Skill.dat` and checked in rather than put in a column: it is static client data, identical for every server, and it reads straight out of the file that ships with the game. 68 skills, 1002 cells.

### Tests

Eight, with the expected numbers taken from the file rather than from another implementation: the two example patterns, the four cardinals, a diagonal, and the two cases that fail **in silence** — a skill with no pattern, and caster and target on the same cell (where normalising divides by zero, every cell becomes `NaN` and then zero, and the whole pattern stacks on the caster without a word).

Zero warnings, all tests pass. This is the first slice of the BCard/combat work from the roadmap’s W 1.0.7 — the rest is larger and will come separately so each piece stays reviewable.

### Reading the CELL section

Two things that are easy to get wrong, moved here out of the code now that the comments are gone:

- A pattern is triples `(dx, dy, continues)`, padded to thirty slots. The third field says whether the list goes on, so a pattern of N cells is N triples ending in 1 followed by zeros. **A pattern that fills every slot ends on a 1 with nothing after it** — that means "from here it is not said", not "there is more". Skill 1175 is that case: thirty cells forming a closed figure, and its COST tail is zeros. The tail has to be read and found empty rather than assumed to hold something. `AFullCellWithAnEmptyTailStopsAtThirty` is the test.
- `SkillCells.Parse` returns null on a malformed value instead of throwing: the pattern decides who a skill hits, and a parse failure should cost the pattern, not the cast.

## Summary by CodeRabbit

* **New Features**
  * Added support for skill-specific cell patterns to define precise hit areas.
  * Area-of-effect targeting now respects custom patterns for monsters and players.
  * Skill data now includes optional cell-pattern information parsed from game data.
* **Bug Fixes**
  * Malformed or incomplete patterns are safely ignored instead of causing errors.
  * Skills without custom patterns continue using their existing rectangular range behavior.
* **Documentation**
  * Documented the new cell-pattern skill data field.


erwan-joly:
Holding this one on placement rather than on the logic — `TargetResolver` and the pattern handling look right.

`SkillCellTable.g.cs` is 85 lines of client data checked into `NosCore.GameObject`, and `CELL` is a `Skill.dat` section. Everything else in that file already goes through `NosCore.Parser` into the database — `CELL` appears in `SkillParser.cs` only as a commented-out sample line, so it is unparsed rather than deliberately excluded.

Two problems with the table as a table:
- it cannot be regenerated from a client update without whatever produced it, which is not in the repo
- the header says the pattern spills into "the unused tail of COST" when it exceeds thirty cells, which is exactly the kind of layout knowledge the parser exists to hold

The `add-dat-field` route would put the pattern on the skill row and delete the generated file. Same behaviour, and it survives the next client patch.

Happy to be wrong if there is a reason CELL cannot be a column — if the pattern genuinely is not per-skill-row, say so and I will merge as is.

denislauri1999:
Taken, on both counts. `SkillCellTable.g.cs` is gone and `CELL` now goes through `NosCore.Parser` onto a `CellPattern` column, like every other section of the file.

**To answer the question you left open — is the pattern per-skill-row?** Yes, and it isn't a judgement call: `Skill.dat` has 1958 `VNUM` sections and 1958 `CELL` sections, one each, exactly like `NAME` and `TYPE`. Sixty-eight are non-empty.

**The layout knowledge is now where you said it belonged.** `CELL` is ninety-three fields: two of header, then triples `(dx, dy, continues)`, ending at the first triple whose third field is zero. 2 + 30×3 + 1 = 93, so the section holds thirty cells at most, and a pattern needing more continues in the tail of `COST` — thirty-three fields of which three are read, zero for 1948 of the 1958.

And the trap inside that, which is why the tail can't simply be appended: **running out of room is not the same as continuing.** A `continues` of 1 on the *last available* triple means "from here it is not said", not "there is more". Skill 1175 fills all thirty with a closed figure and stops; 1857 fills them and goes on. There's a parser test for each.

The column is `"dx,dy,dx,dy,..."`. Not a child table because the pattern is one value of the skill, read whole or not at all — nothing ever asks for its fourth cell; the longest is eighty numbers. A malformed value parses to null rather than throwing, so one bad row degrades a cast to a single target instead of taking it down.

The tests that checked the decoded data moved to `SkillParserTests` against a synthetic `Skill.dat`. What's left in `GameObject` is reading the column and rotating the pattern towards the target.

Zero build warnings; `NosCore.Parser.Tests` 105/105 (5 new), `NosCore.GameObject.Tests` 388/388. The `Skill.dat` documentation snapshot was regenerated for the new column. No play test — I don't start the servers.

denislauri1999:
Rebased and trimmed to the CLAUDE.md that landed today: comments cut back to the ones that answer a question a reader would otherwise have to dig for, and nothing outside the project named. No behaviour change in this push.

src/NosCore.GameObject/Services/BattleService/Model/SkillInfo.cs
```diff
@@ -27,7 +27,10 @@ public sealed record SkillInfo(
     byte Element,
     short Duration,
     short MpCost,
-    IReadOnlyList<BCardDto> BCards)
+    IReadOnlyList<BCardDto> BCards,
+    // The cells this skill hits, from Skill.dat's CELL section via the skill row.
```
erwan-joly: useless comment

src/NosCore.GameObject/Services/BattleService/TargetResolver.cs
```diff
@@ -35,6 +35,13 @@ public IReadOnlyList<IAliveEntity> Resolve(IAliveEntity attacker, IAliveEntity p
             return results;
         }
 
+        // A CELL pattern comes with radius zero: the area is the pattern.
```
erwan-joly: useless comment

src/NosCore.GameObject/Services/BattleService/TargetResolver.cs
```diff
@@ -86,6 +93,13 @@ private static bool IsEnemy(IAliveEntity attacker, IAliveEntity candidate)
         };
     }
 
+    // The pattern's cells when the skill has one, otherwise the box around the target.
```
erwan-joly: useless comment

src/NosCore.GameObject/Services/BattleService/SkillCells.cs
```diff
@@ -0,0 +1,93 @@
...
+// | | ' | \/ |`._`.| \_| \/ | v / _|
+// |_|\__|\__/ |___/ \__/\__/|_|_\___|
+//
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+
+namespace NosCore.GameObject.Services.BattleService;
+
```
erwan-joly, line 12: comment is having some data that is just example not sure how useful those are

denislauri1999: All gone in 2f5177d5 — the four you named plus the file header and the two in the tests, so the whole PR is comment-free now.

The two things that were genuinely non-obvious moved into the PR description instead: the CELL triples end on a `continues` of 1 with nothing after when the pattern fills all thirty slots (that means "from here it is not said", and skill 1175 is the case), and `Parse` returns null on a malformed value rather than throwing.

Build clean, 1004 tests green. `stile-upstream` now reports 0 on this branch.

erwan-joly:
we don't use merge master into pr we use rebase please remove that merge commit and rebase

src/NosCore.GameObject/Services/BattleService/Model/SkillInfo.cs
denislauri1999: Gone. `SkillInfo` now only gains the `string? CellPattern = null` parameter, with no comment on it.

The whole branch is comment-free: `stile-upstream` reports 0 added comment lines against `origin/master`. What the comments used to say is in the PR description instead.

Rebased onto master (`3cfbde3`) so the history is linear again — the merge commit is gone. Build clean at 0 warnings, 1070 tests green with the CI filter (`TestCategory!=OPTIONAL-TEST`).

src/NosCore.GameObject/Services/BattleService/TargetResolver.cs
denislauri1999: Gone. `TargetResolver` now carries no comments at all — the only additions are the `SkillCells.Parse`/`Resolve` call and the `IsHit` helper that replaces the two `WithinRange` call sites.

`stile-upstream` reports 0 added comment lines on the branch against `origin/master`.

Rebased onto master (`3cfbde3`), build clean at 0 warnings, 1070 tests green with the CI filter.

src/NosCore.GameObject/Services/BattleService/TargetResolver.cs
denislauri1999: Gone. `TargetResolver` now carries no comments at all — the only additions are the `SkillCells.Parse`/`Resolve` call and the `IsHit` helper that replaces the two `WithinRange` call sites.

`stile-upstream` reports 0 added comment lines on the branch against `origin/master`.

Rebased onto master (`3cfbde3`), build clean at 0 warnings, 1070 tests green with the CI filter.

PR #2292, opened 2026-08-23
Import the time-space and raid entrances

denislauri1999:
First slice of #2282, split as you asked. This one does not touch the definition format at all, so it can be judged on its own while that question is still open.

`ScriptedInstance` was an `IEntity` that nothing ever filled: the table existed and the import never wrote a row, so no entrance had a level range or a heroic flag for anything to read.

It becomes an `IStaticEntity` with `LevelMinimum`, `LevelMaximum` and `IsHeroic`, and a parser reads the entrances out of the capture — `wp` for the time-spaces, `rbr` for the raids, all three raid portal types.

7 tests over the parser. Build clean, everything green.

The rest of #2282 splits into two more: the minimap markers and entry panel, which are also format-independent, and then the run/room machinery on whichever format you settle on. I have not opened those yet — the second one waits on your call between C# definitions and JSON, and on what should happen to the existing XML content.

## Summary by CodeRabbit

- **New Features**
  - Added automatic parsing and importing of scripted time-space instances from game data.
  - Imported instances now include level requirements, heroic classification, entrances, and waypoint details.
  - Full and prompted imports can include scripted instances.

- **Improvements**
  - Added localized messages reporting parsed instance counts.
  - Improved duplicate handling and filtering of unknown map data.
  - Existing instance metadata is now refreshed during imports.

### DIFF
--- a/src/NosCore.Data/Enumerations/I18N/LanguageKey.cs
+++ b/src/NosCore.Data/Enumerations/I18N/LanguageKey.cs
@@ -65,6 +65,7 @@ namespace NosCore.Data.Enumerations.I18N
         ITEMS_PARSED,
         MAPS_PARSED,
         PORTALS_PARSED,
+        TIMESPACES_PARSED,
         MAPS_LOADED,
         NO_MAP,
         MAPMONSTERS_LOADED,
--- a/src/NosCore.Data/Resource/LocalizedResources.cs.resx
+++ b/src/NosCore.Data/Resource/LocalizedResources.cs.resx
@@ -554,4 +554,7 @@
   <data name="UNHANDLED_UPGRADE_TYPE" xml:space="preserve">
     <value>Neznámý typ vylepšení: {0}</value>
   </data>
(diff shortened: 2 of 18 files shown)


denislauri1999:
Both fixed, and the second one was worth catching.

**Backfilling the metadata** — right, and it defeated the whole PR on any database that already holds entrances. The columns arrive from the migration as `0`, `0` and `false`, and a stored entrance was skipped outright, so an upgraded database would have kept every level range at zero and every raid non-heroic, silently.

A stored row now has its metadata refreshed. It keeps its `ScriptedInstanceId`, so the upsert updates rather than inserting a duplicate, and it keeps its `Script`, which is not this parser's to write.

The in-run dedup is now separate and still there — the capture walks a map more than once and the same `wp` comes round again. There is a test for that case too, since splitting the two is exactly where it could regress.

The test that covered this was pinning the wrong behaviour: it asserted nothing was saved. Rewritten to assert the metadata changes and the script survives — under the old code it fails.

**Spanish** — taken, `instancias de script`.

110 parser tests, build clean.

PR #2302, opened 2026-08-24
The Adventurer's skill list, and the rows a class change leaves behind

denislauri1999:
Three things in `SkillService`, all about which skills a character is allowed to hold.

### Class 0 is not the Adventurer

It is the scrap container of `Skill.dat`: 193 entries end up there — passives, monster skills, things with no cost and no cast time. `s.Class == (byte)character.Class` therefore gave an Adventurer *all* of them, and the bar filled with icons the client will not cast. The symptom is "the skills arrive but they do not work".

The Adventurer's ten real skills are now listed by number. 209 does not exist.

### A class change was only half done

The change empties the in-memory list and learns the new class's skills. Nothing ever deleted the rows behind the old ones, so the next login loaded both sets back.

That is not a cosmetic leftover. Cast ids are numbered per class and start at zero, so an Archer who used to be an Adventurer knew two skills answering to cast 0:

| skill | class | cast id | profile |
|---|---|---|---|
| Swing | Adventurer | 0 | melee |
| Archery | Archer | 0 | ranged |

Which one the resolver returned came down to dictionary order. The visible symptom was a basic attack computed off the **wrong weapon**: a melee skill selects the secondary-weapon profile on an Archer, so the bow in the main hand counted for nothing.

The cleanup also runs from `LoadSkill`, so characters already in that state heal themselves at their next login rather than staying broken until they change class again.

### LearnClassSkillsAsync inserted a duplicate row on every call

`Id = Guid.NewGuid()` each time. In memory nothing showed — the dictionary is keyed by skill number and collapses them — but the rows piled up behind it. The existing row's id is reused when there is one.

---

The specialist card's skills are a separate change and are not in this PR.

## Summary by CodeRabbit

- **New Features**
  - Class changes now automatically relearn skills available to the character’s new class and job level.
  - Adventurer skill selection now includes the correct capture skill and excludes invalid or unrelated skills.

- **Bug Fixes**
  - Skills unavailable due to class or job-level requirements are removed from active and saved character data.
  - Relearning skills no longer creates duplicate entries.

- **Tests**
  - Added coverage for class filtering, job-level requirements, invalid skills, and repeated skill learning.

### DIFF
--- a/src/NosCore.GameObject/Ecs/Extensions/PlayerBundleExtensions.cs
+++ b/src/NosCore.GameObject/Ecs/Extensions/PlayerBundleExtensions.cs
@@ -945,7 +945,7 @@ public static class ClientSessionMailExtensions
     public static async Task ChangeClassAsync(this ClientSession session, CharacterClassType classType,
         IOptions<WorldConfiguration> worldConfiguration,
         IExperienceService experienceService, IJobExperienceService jobExperienceService, IHeroExperienceService heroExperienceService,
-        IItemGenerationService itemProvider)
+        IItemGenerationService itemProvider, Services.SkillService.ISkillService skillService)
     {
         var character = session.Character;
         var inventoryService = character.InventoryService;
@@ -982,6 +982,18 @@ public static class ClientSessionMailExtensions
         character.Class = classType;
         character.Hp = character.MaxHp;
         character.Mp = character.MaxMp;
+
+        // The old class's skills are no longer usable and the new one's are not there yet:
+        // without this you change job and keep the previous bar, full of icons the client
+        // refuses to cast because they do not belong to the class.
+        character.Skills.Clear();
+
+        // Emptying the list is not enough - the rows behind it survive, and the next login loads
+        // them straight back on top of the new class's. LearnClassSkillsAsync then grants what
+        // the job level allows, and the change has just put that back to 1, so it starts from
+        // the first skill.
+        await skillService.ForgetSkillsOfOtherClassesAsync(character).ConfigureAwait(false);
+        await skillService.LearnClassSkillsAsync(character).ConfigureAwait(false);
 
         var itemsToAdd = worldConfiguration.Value.BasicEquipments.TryGetValue(classType.ToString(), out var byOrigin)
             && byOrigin.TryGetValue(StarterOrigin.CreateAndUpgrade, out var pack)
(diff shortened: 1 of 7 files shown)


denislauri1999:
Pushed a second commit: the cleanup now also runs where the stale rows are created, not only at the next login.

Looking at it again, `ChangeClassAsync` on master does not touch `Skills` at all — the character keeps the old class's bar, in memory and in the database, and never learns the new class's. So the first commit repaired characters already in that state at their next login, but left the window in between. This closes it: clear the list, delete the rows behind it, learn what the new job level allows (the change has just put it back to 1, so it starts from the first skill).

`ISkillService` is threaded through the two callers.

denislauri1999:
Both findings taken, and checking the first one against `Skill.dat` turned up a worse mistake of my own.

**209 is not a hole in the numbering — it is the Adventurer's Capture.** My list left it out and the comment claimed it does not exist. The file says class 0, cast id 16, `LevelMinimum` 1, name Capture. That is why `CharNewPacketHandler` grants it to every new character, and why the comment there says a missing one makes `u_s 16` answer `cancel 2` — 16 is exactly this skill's cast id.

As pushed, the cleanup deleted 209 from every Adventurer, and nothing would have reported it: they just stop being able to catch pets. The test asserted the omission and passed. It now asserts the opposite and fails if 209 leaves the list.

211 and 212 stay out on purpose — the file calls them "Ultra Super Cheating Skill" and "Admin Cheating Skill".

Written down rather than guessed at: 300–306 are also class 0, `LevelMinimum` 10–18, names like "Strengthen Swing". They look like upgraded forms of 200–206 and they carry the **same cast ids**, so they cannot simply be added alongside. Nothing in the files says how one replaces the other, so they stay out until something does.

**The job-level filter is right.** A class change resets the job level to 1, so a persisted row for a destination-class skill needing job 20 is as unusable as one belonging to the class left behind, and `LoadSkill` restored it past any level check. Applied. I also renamed the method to `ForgetUnlearnableSkillsAsync` — with that filter it no longer only forgets *other classes*, and the old name would have lied.

**The duplicate-row test was weak, as you say.** It counted the in-memory dictionary, which is keyed by skill number and collapses duplicates — it could not see the thing it defended. It now uses a store that remembers, clears the in-memory list the way a relog does, and asserts the row id is reused.

PR #2304, opened 2026-08-24
The cards a skill inflicts, which nothing read

denislauri1999:
BCard type 25 is the most widespread effect in the game — **1344 skills declare one** — and nothing reads it.

A skill does not carry the effect. It carries a BCard saying *"with N% chance, apply Card number M"*, and M is a real entry of `Card.dat` with its own duration and its own BCards:

```
11: Has a %s%% probability of causing [%s].
12: There is a %s%% chance that %s will be removed.
```

### Which field is which

`FirstData` is the percentage, `SecondData` is the card id. The names settle it:

| skill | prob. | card |
|---|---:|---|
| Star Attack | 60% | 7 — *Blackout* |
| Hit of Rage | 2% | 4 — *Anger* |
| Blood Oath | 100% | 17 — *Blood Oath* |
| Suppress | 100% | 19 — *Suppressed* |

1340 of the 1341 ids a skill names exist in `Card.dat`.

Worth stating, because reading it the other way round is easy and quiet: with 2759 cards spread over ids 0 to 4440, `FirstData` is **also** a valid card id 1278 times out of 1341. Checking whether the number exists gives the wrong answer with a crushing majority. What separates them is the shape of the columns — 32 distinct values on one side, 780 on the other. An id does not repeat 717 times; a probability does.

### Who receives the card is *not* in the BCard

Battle Cry declares "100% of card Battle Cry" and is a self-buff. Suppress declares its card with exactly the same structure and is a debuff on the enemy. The difference lives in the skill's `TARGET` section, which the file writes as bare numbers with no sentence explaining them — so the files cannot answer it, and I did not guess.

This therefore hangs off **a blow that has landed**, where the question does not arise: the entity that took the damage is the one the card goes on. Skills that damage nobody never reach the path, so self-buffs are outside it *by construction rather than by omission* — that is 704 of the 1341 declarations, the ones whose `TargetType` is 0. The rest needs a capture of two casts to settle; the `bf` packet names the entity that receives the buff.

### Notes for review

- `BuffService.ApplyAsync` already did the right thing, `bf` packet included, and had **no callers**. Only `ApplySkillBuffAsync` was used, which turns the skill's *own* BCards into a lasting effect — a different thing.
- `HitQueue.TryApplyHit` becomes `async`. It was returning `Task.CompletedTask` throughout; the effect a blow carries has to follow it, not race it.
- **`CardCatalog` / `ICardCatalog` also appear in #2293**, where `EquipmentStatsService` needed them first. They are self-contained — two lists that the container already registers, because `PersistenceModule` registers a `List<T>` for every `IStaticEntity`. Whichever of the two PRs lands first, the other rebases and the duplicate disappears.
- Swapping the two fields fails three of the seven tests.

### Reading the declaration

Moved here out of the code now that the comments are gone:

- **`FirstData` is the percentage and `SecondData` the card id**, and the columns are what tell them apart: 32 distinct values on one side against 780 on the other, and an id does not repeat 717 times. Getting them the wrong way round raises nothing — `FirstData` is a valid card id in 1278 of the 1341 declarations, so a server that swapped them would apply a real, wrong card at a real, wrong probability and never complain. The test that catches it is `TheCardTheSkillNamesIsTheOneApplied`.
- **717 of the 1341 declarations say 100.** `Next(0, 100)` yields 0 to 99, so the comparison has to be "less than": 100 always succeeds, 0 never. An off-by-one there would silently drop the most common case of all.
- **Subtype 11 causes and subtype 12 removes** — opposite actions, so treating them alike would apply the card a skill is meant to strip off.
- **One of the 1341 ids is not in `Card.dat`.** It is skipped rather than thrown on: a single bad row in the client's data must not take a blow down with it.

## Summary by CodeRabbit

* **New Features**
  * Added percentage-based damage effects tied to maximum health.
  * Added support for applying and removing card-based buffs during combat.
  * Special actions such as push, pull, and charge now occur before hit credit.
  * Vitality refreshes automatically after relevant damage and buff changes.

* **Bug Fixes**
  * Corrected handling of self-inflicted effects and multiple percentage-damage effects.
  * Ensured combat actions wait for buff application to complete.


denislauri1999:
Taken — and it was the same shape of hole this PR is about, one level up.

`InflictedCardService` had seven tests; the step that reaches it had none. Both HitQueue cases passed an anonymous `IInflictedCardService` mock, so removing the call, or inverting its condition, broke nothing.

Named the mock in both:

- **landed hit** — asserts `InflictAsync` ran once, with the target and the skill's BCards
- **killing hit** — asserts it never ran. Poisoning a corpse costs a packet and a buff icon on an entity that is about to stop existing, and it matches the guard the existing buff application already had.

Checked by breaking it, since that is the only thing that proves a test: removing the call fails one, inverting `!killed` fails two.

denislauri1999:
Right again, and it is the sharper version of the last one: a `Verify` passes whether the call is awaited or fired and forgotten, because Moq answers with a completed `Task` either way. So the test defended *that the call happens* and not the one property `TryApplyHit` was made `async` for — and fire-and-forget is exactly what this used to be.

Added a separate test: the mock returns a task that stays pending, and the hit must not report itself finished while it does.

I did not assert the ordering with a bare "is it complete yet", which would race the worker. It waits, and the wait can only fail in the safe direction: if the call is awaited the hit can never complete, so the delay always wins. A loaded machine can make this pass when it should not, never fail when it should not.

Putting the `_ =` back fails it.

denislauri1999:
Both taken, and the first one is the sharper of the three so far.

**The signal.** The test leant on a delay alone, so it also passed on a machine slow enough that the worker never reached the call — a pass for the wrong reason, which is exactly the failure mode this PR exists to fix. The mock now signals on entry, the test waits for that signal, and only then asserts the hit is still unfinished. The remaining delay covers only the negative half, where it can fail in the safe direction.

**The finally.** Right, and worse than it looks: a failed assertion left the queue's worker parked on a `TaskCompletionSource` nobody would ever complete, for the rest of the run. The gate is released in a `finally` now.

Putting the `_ =` back still fails it.

src/NosCore.GameObject/Services/BattleService/InflictedCardService.cs
```diff
@@ -0,0 +1,76 @@
...
+
+namespace NosCore.GameObject.Services.BattleService;
+
+// BCard type 25/11 and 25/12: FirstData is the percentage, SecondData the card id. The columns
+// tell them apart - 32 distinct values on one side against 780 on the other, and an id does not
+// repeat 717 times.
+//
+// Takes a target rather than a skill because the BCard does not say who receives the card; on a
+// blow that has landed the entity that took the damage is the one it goes on.
+public sealed class InflictedCardService(
```
erwan-joly: likely the code here should be in IBuffService

src/NosCore.GameObject/Services/BattleService/InflictedCardService.cs
```diff
@@ -0,0 +1,76 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using NosCore.Data.Enumerations.Buff;
+using NosCore.Data.StaticEntities;
+using NosCore.GameObject.Ecs.Interfaces;
+using NosCore.GameObject.Infastructure;
+
+namespace NosCore.GameObject.Services.BattleService;
+
+// BCard type 25/11 and 25/12: FirstData is the percentage, SecondData the card id. The columns
```
erwan-joly: don't need example of which bcard number in the comments. Comments overall should only be to explain code that is too hard and in most of cases we shouldnt have hard code

denislauri1999: Both done in 0cc3bd1b.

`InflictedCardService` is gone — its loop is now `BuffService.InflictCardsAsync`, next to `ApplySkillBuffAsync` which is the other skill-originated entry point. That drops an interface, a class and a constructor parameter on `HitQueue`, and the two calls it made are local now.

One consequence worth naming: the tests used to verify calls on a mocked `IBuffService`, which is not available once the code IS `IBuffService`. They assert the real thing now, through `GetActiveBuffs`/`HasBuff` on a `MonsterComponentBundle` carrying only the two components the buff path reads. I measured red first by swapping FirstData and SecondData in the loop: 3 failed of 7.

Comments all gone, here and in the tests, and `stile-upstream` reports 0 on the branch. What they said — FirstData is the percentage and SecondData the id, 717 of 1341 declarations say 100 so the comparison must be "less than", subtype 12 is the opposite action, and one declared id is not in Card.dat — is in the PR description.

Build clean, 999 tests green.

src/NosCore.GameObject/Services/BattleService/HitQueue.cs
```diff
@@ -183,6 +183,13 @@ await ApplySpecialActionsAsync(request.Skill.BCards, request.Origin, target)
                 _ = buffService.ApplySkillBuffAsync(target, request.Skill.SkillVnum, request.Skill.Duration, request.Skill.BCards, request.Origin);
             }
 
+            if (!killed && request.Skill.BCards.Count > 0)
+            {
+                await buffService
+                    .InflictCardsAsync(target, request.Origin, request.Skill.BCards)
+                    .ConfigureAwait(false);
```
erwan-joly: we should remove all those they are not useful in our case

erwan-joly:
need rebase

src/NosCore.GameObject/Services/BattleService/HitQueue.cs
denislauri1999: Removed in d1aca026 — all three `ConfigureAwait(false)` this PR added, same reason you gave on #2321.

Same push also merges master in and ports `InflictCardsAsync` to the `BCardEffect` key, since #2325 deleted `AdditionalTypes` and the branch no longer compiled against master. Build clean, 1021 tests green with the CI filter.

PR #2306, opened 2026-08-24
Damage as a percentage of HP, which the case did not read

denislauri1999:
BCard type 37 subtype 31 — *"Decreases the opponent's HP by %s%%."*

**112 of the 122 declarations on skills are this one subtype**, and nothing read it: the skills carrying it did none of what they declare. The percentages in the file run from 5% to 100%, with 90% appearing twelve times — these are boss skills meant to take most of a health bar in one blow.

### One assumption is ours

The percentage comes off **maximum** HP. The file says only "HP by %s%%" and does not distinguish, so this is the one point that does not come from it. Off current HP the loss would halve and halve again without ever finishing anything, which is not what a skill declaring 90% is for.

It is the same assumption the sibling codebase already carries, so the two do not drift.

### One difference from the sibling, on purpose

It can kill. The loss is part of the same subtraction as the blow, so the ordinary death path handles it.

The sibling floors the victim at 1 HP, but that is a workaround for *where* it applies the effect — outside the blow, where a kill would bypass the death sequence. Here that does not arise, and nothing in the file says the effect must leave its victim standing.

### Left out

Subtype 32 hits the caster instead, and no skill in the file declares it. Written blind against no data it would be a guess, so it is not here.

### Checked by breaking it

Reading the percentage off current HP instead of maximum fails two of the five tests. One test also documents an existing guard I ran into: a blow whose ordinary damage rolls to zero is already treated as a miss, so the percentage rides along with a blow that landed rather than standing on its own.

### DIFF
--- a/src/NosCore.GameObject/Services/BattleService/HitQueue.cs
+++ b/src/NosCore.GameObject/Services/BattleService/HitQueue.cs
@@ -4,9 +4,6 @@
 // |_|\__|\__/ |___/ \__/\__/|_|_\___|
 //
 
-using NosCore.Data.Enumerations.Buff;
-using NosCore.Data.StaticEntities;
-using System.Collections.Generic;
 using System;
 using System.Collections.Generic;
 using System.Collections.Concurrent;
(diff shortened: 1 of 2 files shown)


denislauri1999:

`HitQueue` gains an `IInflictedCardService` parameter in #2304. The `QueueDealing` helper this PR adds to `HitQueueTests` constructs it with the current five arguments, so after both land it needs `new Mock<IInflictedCardService>().Object` before the logger.

PR #2307, opened 2026-08-24
Elemental resistance, which the fold had no case for

denislauri1999:
BCard type 13, *"Changes elemental resistance"* — **561 declarations**, 408 on items and 153 on cards — and the buff fold had no case for it. A resistance buff changed nothing.

Almost everything needed was already there: the four fields exist on `CombatStats` and are already read in `ComputeElementalDamage` as a percentage taken off the incoming elemental damage. Only the step that fills them was missing.

```
11: All elemental resistance is increased by %s.
21: Fire resistance is increased by %s.
31: Water resistance is increased by %s.
41: Light resistance is increased by %s.
51: Shadow resistance is increased by %s.
```

Not to be confused with type 7 in the case just below it: that is the **attacker's** element rate, this is the **defender's** resistance to it.

Two notes:

- **"All" adds to each of the four** rather than living in a fifth field. The damage step reads one resistance, chosen by the attacker's element, so a separate total would have to be remembered at every one of those reads.
- Unlike the types where the file repeats the same sentence in both slots, this one really does pair increase with decrease — *"increased by %s"* against *"decreased by %s"* — so X2 subtracts.

Writing it on the buff fold rather than beside the equipment means the item half arrives with no further work, since the worn pieces' BCards go through the same place.

### Checked by breaking it

"All" not reaching every element fails two tests; reading the decrease as an increase fails one.

### DIFF
--- a/src/NosCore.GameObject/Services/BattleService/BattleStatsProvider.cs
+++ b/src/NosCore.GameObject/Services/BattleService/BattleStatsProvider.cs
@@ -172,6 +172,7 @@ public sealed class BattleStatsProvider(IBuffService buffService) : IBattleStats
         int hitRateFlat = 0, dodgeFlat = 0;
         int moraleFlat = 0;
         int elementAll = 0, elementFire = 0, elementWater = 0, elementLight = 0, elementDark = 0;
+        int resistAll = 0, resistFire = 0, resistWater = 0, resistLight = 0, resistDark = 0;
 
         foreach (var buff in buffs)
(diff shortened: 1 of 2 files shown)


denislauri1999:

`BattleStatsProvider` gains an `IEquipmentStatsService` parameter in #2293. The tests this PR adds construct it with one argument, so after both land they need `new Mock<IEquipmentStatsService>().Object` as a second.

PR #2308, opened 2026-08-24
What the attacker does to the defender's resistances

denislauri1999:
BCard type 14, *"Changes enemy's resistances"* — **210 declarations**, and the mirror of type 13 in #2307. That one is the resistance of whoever is being hit; this is what the one hitting does to it, so it rides on the attacker and is read against the defender's own number.

```
11: Increases the enemy's elemental resistances by %s.
12: Reduces the enemy's elemental resistances by %s.
21..52: the same, one element at a time.
```

A *"reduces by 20"* folds to `-20`, so the damage step simply adds the two. The sign is the whole point: read backwards, resistance-piercing effects would protect the target instead, and nothing would report it — the fight goes on and the numbers are just wrong.

Written on the fold rather than beside the equipment, so the 194 declarations on items arrive with no further work once the worn pieces' BCards reach the same place.

### The tests needed a second layer

Four of the seven go through the damage step with the field set by hand. A mutation that folded the *water* value into the *fire* field left all four green — so three more go through `BattleStatsProvider` itself. Testing the new code and not the step that fills it is a hole I had pointed out to me on #2304 earlier today.

Flipping the sign fails two; folding the wrong element fails one.

### One known conflict, and it is trivial

This conflicts with **#2305** on `CombatStats.cs`, and only there. Both PRs add optional parameters to the same positional record, and C# requires those to come last — so both insertions land at the same point and git cannot merge them. The resolution is *keep both blocks*; nothing has to be decided.

I checked every other pair among my open PRs with `git merge-tree` (36 combinations) and this is the only one left.

## Summary by CodeRabbit

* **New Features**
  * Elemental resistance increases and reductions from active buffs now affect combat damage calculations.
  * Added support for fire, water, light, dark, and all-element resistance modifiers.
  * Resistance modifiers are applied only to matching attack elements.

* **Bug Fixes**
  * Improved handling of resistance-piercing and resistance-increasing effects, including zero-value and sign behavior.

### DIFF
--- a/src/NosCore.GameObject/Services/BattleService/BattleStatsProvider.cs
+++ b/src/NosCore.GameObject/Services/BattleService/BattleStatsProvider.cs
@@ -170,6 +170,7 @@ public sealed class BattleStatsProvider(IBuffService buffService) : IBattleStats
         int critInflicting = 0, critDamage = 0;
         int defenceAll = 0, defenceMelee = 0, defenceRanged = 0, defenceMagical = 0;
         int hitRateFlat = 0, dodgeFlat = 0;
+        int foeAll = 0, foeFire = 0, foeWater = 0, foeLight = 0, foeDark = 0;
         int moraleFlat = 0;
         int elementAll = 0, elementFire = 0, elementWater = 0, elementLight = 0, elementDark = 0;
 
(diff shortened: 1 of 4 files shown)


denislauri1999:

`BattleStatsProvider` gains an `IEquipmentStatsService` parameter in #2293. The tests this PR adds construct it with one argument, so after both land they need `new Mock<IEquipmentStatsService>().Object` as a second.

PR #2317, opened 2026-08-26
The two BCard types the files do name

denislauri1999:
## What

Types 93 and 94 keep a number while their names resolve in the client's own language files:

```
93  Changes the elemental percentage.
94  Changes the elemental resistance percentage.
```

So they become `ElementPercentage` and `ElementResistancePercentage`, and their members follow.

## Why these two are worth naming

They are the **percentage halves of types 7 and 13** — the same quantity declared in a second unit. And the pairing is not by number:

| | 11 | 21 | 31 | 41 | 51 |
|---|---|---|---|---|---|
| 7 and **93** — element | fire | water | light | shadow | **all** |
| 13 and **94** — resistance | **all** | fire | water | light | shadow |

93 follows 7's subtype order and 94 follows 13's. A reader looking at two bare numbers has no way to know that; a reader looking at the names does.

## Why the neighbours keep their numbers

The types around them keep a number for a reason worth stating: **their own name token resolves in none of the eight languages the client ships.** They are buckets by content epoch rather than by mechanic, so a name would describe at most half of what is inside.

## Testing

- Builds with **0 warnings**; full suite green; `NosCore.Parser.Tests` passes 104/104 with `TestCategory!=OPTIONAL-TEST`.
- Rebuilt on top of the flat `BCardEffect` enum — the earlier version of this PR was written against `AdditionalTypes.cs` and no longer applied.
- A rename only. Nothing reads these yet.


denislauri1999:
Rebased and trimmed to the CLAUDE.md that landed today: comments cut back to the ones that answer a question a reader would otherwise have to dig for, and nothing outside the project named. No behaviour change in this push.

PR #2320, opened 2026-08-27
A cast could put a character inside a wall

denislauri1999:
`u_s` carries a position, and `UpdateAttackerPosition` wrote it straight into `PositionX`/`PositionY`. `WalkPacketHandler` refuses a destination that is not walkable; this path had no such check, so the same movement was allowed by one road and refused by the other.

The guard is the walk path's own `IsWalkable`, so the two agree by construction rather than by two copies of a rule.

**Tested:** `NosCore.GameObject.Tests` 442/442 and `NosCore.PacketHandlers.Tests` 410/410, solution builds with zero warnings. Not played - I do not start the servers.

## Summary by CodeRabbit

* **Bug Fixes**
  * Characters are no longer moved onto non-walkable tiles during battle positioning.
  * Invalid or unavailable destination coordinates now leave the attacker’s position unchanged.

### DIFF
--- a/src/NosCore.GameObject/Services/BattleService/BattleService.cs
+++ b/src/NosCore.GameObject/Services/BattleService/BattleService.cs
@@ -245,8 +245,6 @@ namespace NosCore.GameObject.Services.BattleService
             }
         }
 
-        // WalkPacketHandler refuses a destination that is not walkable; this path wrote the same
-        // two fields with no check, so a cast could put a character inside a wall.
         private static void UpdateAttackerPosition(IAliveEntity origin, HitArguments arguments)
         {
             if (arguments is not { MapX: not null, MapY: not null })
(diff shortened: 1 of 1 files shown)


src/NosCore.GameObject/Services/BattleService/BattleService.cs
```diff
@@ -245,13 +245,22 @@ public async Task TickCooldownResetsAsync(MapInstance mapInstance)
             }
         }
 
+        // WalkPacketHandler refuses a destination that is not walkable; this path wrote the same
```
erwan-joly: useless comment

denislauri1999: Removed in c1585836 — the comment is gone, the guard itself is unchanged.

erwan-joly (approved):

PR #2321, opened 2026-08-27
Maximum HP and MP stop being a login-time snapshot

denislauri1999:
Maximum HP and MP were computed once, at login, from class and level. Two things counted for nothing:

- the HP and MP bonus of worn items — the parser has always read `Item.Hp`, and nothing consumed it;
- **effect type 33**, which sits on 706 items and is the second most widespread effect in the data.

A thousand-HP armour gave zero, and nothing reported it: the bar the player saw was the right one for their level, so the defect was invisible.

`VitalityService` recomputes the maximum from class and level, plus equipment, plus active effects, and notifies. It runs where the inputs change: on wear/unwear, and after a skill's buff has been applied.

**The order is read from the files, not chosen.** `BCard.dat` type 33 subtypes 31 and 41 (`Increases maximum HP/MP by %s%%`) add "(Only used by buffs.)", which settles it: the percentages apply to the total that already includes the equipment, not to the base alone. Applying them first would give lower numbers, wrong in a way nothing reports.

One await changed in `HitQueue`: the buff application is now awaited before the recomputation, so a type 33 buff that has just landed is visible to it. The worker already serialises per target, so this orders work that was already happening rather than adding any.

**Tested:** `NosCore.GameObject.Tests` 455/455 (13 new, covering the flat and percentage subtypes and the order between them), `NosCore.PacketHandlers.Tests` 410/410, solution builds with zero warnings. Not played — I do not start the servers.

## Summary by CodeRabbit

* **New Features**
  * Character maximum HP and MP now accurately reflect level, class, equipment, and active effects.
  * Vitality updates are automatically applied after equipment changes, leveling up, combat effects, character selection, and expired buffs.
  * Current HP and MP are clamped to updated maximum values, with changes communicated to the client.

* **Bug Fixes**
  * Corrected missing health and mana updates caused by equipment and temporary effects.

* **Tests**
  * Added coverage for vitality calculations, bonuses, caps, level changes, and health clamping.

### DIFF
--- a/src/NosCore.GameObject/Services/BattleService/HitQueue.cs
+++ b/src/NosCore.GameObject/Services/BattleService/HitQueue.cs
@@ -60,8 +60,6 @@ public sealed class HitQueue(
         return request.Completion.Task;
     }
 
-    // A buff can move maximum HP and MP (BCard type 33), and the maximum does not live in
-    // CombatStats: nothing else would pick the change up.
     private async Task RefreshVitalityAsync(IAliveEntity entity)
     {
         if (entity is ICharacterEntity character)
@@ -115,7 +113,7 @@ public sealed class HitQueue(
     }
 
     // async because the effects a blow carries are awaited below: they have to follow the
-    // blow, not race it, and the maximum HP recomputed after them has to see them applied.
+    // blow, not race it.
     private async Task TryApplyHit(HitRequest request)
     {
         try
@@ -186,8 +184,6 @@ public sealed class HitQueue(
                 regenerationService.NotifyDamaged(hurtCharacter.CharacterId);
             }
 
-            // Skill BCards that don't describe damage (i.e. stat modifiers) become a
-            // buff on the target lasting the skill's Duration.
             if (!killed && request.Skill.Duration > 0 && request.Skill.BCards.Count > 0)
             {
                 await buffService
@@ -195,11 +191,6 @@ public sealed class HitQueue(
                         request.Skill.BCards, request.Origin)
                     .ConfigureAwait(false);
 
-                // Awaited and not fire-and-forget any more: the maximum HP below is read from
-                // the effect that has just landed, and a type 33 buff that has not been
-                // applied yet would leave the maximum at its old value until the next piece
-                // of gear changes. The worker already serialises per target, so this only
-                // orders the work that was already happening.
                 await RefreshVitalityAsync(target).ConfigureAwait(false);
                 if (!ReferenceEquals(request.Origin, target))
                 {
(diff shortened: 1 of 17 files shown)


src/NosCore.GameObject/Services/BattleService/HitQueue.cs
```diff
@@ -176,11 +187,24 @@ await ApplySpecialActionsAsync(request.Skill.BCards, request.Origin, target)
...
+            // buff on the target lasting the skill's Duration.
             if (!killed && request.Skill.Duration > 0 && request.Skill.BCards.Count > 0)
             {
-                _ = buffService.ApplySkillBuffAsync(target, request.Skill.SkillVnum, request.Skill.Duration, request.Skill.BCards, request.Origin);
+                await buffService
+                    .ApplySkillBuffAsync(target, request.Skill.SkillVnum, request.Skill.Duration,
+                        request.Skill.BCards, request.Origin)
+                    .ConfigureAwait(false);
+
+                // Awaited and not fire-and-forget any more: the maximum HP below is read from
```
erwan-joly: again too much comments

src/NosCore.GameObject/Services/BattleService/HitQueue.cs
```diff
@@ -176,11 +187,24 @@ await ApplySpecialActionsAsync(request.Skill.BCards, request.Origin, target)
...
-            // buff on the target lasting the skill's Duration. Fire-and-forget is fine:
-            // the worker is already serialising per-target, so ordering is preserved.
+            // buff on the target lasting the skill's Duration.
             if (!killed && request.Skill.Duration > 0 && request.Skill.BCards.Count > 0)
             {
-                _ = buffService.ApplySkillBuffAsync(target, request.Skill.SkillVnum, request.Skill.Duration, request.Skill.BCards, request.Origin);
+                await buffService
+                    .ApplySkillBuffAsync(target, request.Skill.SkillVnum, request.Skill.Duration,
+                        request.Skill.BCards, request.Origin)
+                    .ConfigureAwait(false);
```
erwan-joly, line 192: we don't need the configureawait

src/NosCore.GameObject/Services/BattleService/HitQueue.cs
denislauri1999: Done in 103a9e7a: 89 lines of my comments removed across the PR, and the 9 ConfigureAwait(false) I had added are gone too. What is left is one line per non-obvious spot, nothing more.

Two of your existing comments were touched by my first commit; I put one back word for word (`// blow, not race it.`) and removed the other, because the code now awaits and the sentence said fire-and-forget was fine — a wrong comment is worse than none.

Build clean, 1004 tests green.

PR #2322, opened 2026-08-27
The specialist card's skills, told by the file and not by arithmetic

denislauri1999:
Wearing a specialist card gave its skills by arithmetic: a base vnum plus the morph, which happens to be right for the first twenty-nine cards and wrong afterwards.

`Skill.dat` declares which skills belong to which card. The morph list has gaps - 30, 36, 37, 43, 44 and 50 are absent - so from the first gap onwards the two numberings drift apart and every card past it hands out another card's skills.

The skills now come from the file's own declaration. `ISkillService` gained the lookup; `TransformationService` calls it on wear and clears on remove.

**How the drift was confirmed rather than assumed:** for morphs 1..29 the arithmetic and the file agree, and each card's first skill matches its name - Pyjama/Pillow Fight, Ranger/Archery, Volcano/Magma Ball. Past the first gap they stop matching, and the mismatch is what a player sees: a card showing skills that belong to a different specialist.

**Tested:** `NosCore.GameObject.Tests` 450/450 (new cases for a card below the first gap and one above it), `NosCore.PacketHandlers.Tests` 410/410, zero build warnings. Not played - I do not start the servers.

## Summary by CodeRabbit

- **New Features**
  - Specialist card skills now load automatically when transforming.
  - Specialist skills are filtered by the card’s level and correctly support skills across multiple classes.
  - Swapping or removing a specialist card updates the available skill set immediately.

- **Bug Fixes**
  - Prevented skills from unrelated specialist cards from appearing.
  - Preserved regular class skills during transformations.
  - Removed outdated or unavailable skills when logging in or changing specialist cards.

### DIFF
--- a/src/NosCore.GameObject/Services/SkillService/SkillService.cs
+++ b/src/NosCore.GameObject/Services/SkillService/SkillService.cs
@@ -36,13 +36,13 @@ namespace NosCore.GameObject.Services.SkillService
             await SendSkillListAsync(character, useSpecialist: false);
         }
 
-        private const int FirstSpecialistClass = 31;
+        private const int PyjamaSpecialistClass = 32;
 
         public async Task LoadSpecialistSkillsAsync(ICharacterEntity character, short morph, byte spLevel)
         {
             RemoveSpecialistSkills(character);
 
-            foreach (var skill in skills.Where(s => s.Class > FirstSpecialistClass
+            foreach (var skill in skills.Where(s => s.Class >= PyjamaSpecialistClass
                                                     && s.UpgradeType == morph
                                                     && s.LevelMinimum <= spLevel))
             {
@@ -68,7 +68,7 @@ namespace NosCore.GameObject.Services.SkillService
         private static void RemoveSpecialistSkills(ICharacterEntity character)
         {
             foreach (var entry in character.Skills
-                         .Where(s => s.Value.Skill?.Class >= FirstSpecialistClass)
+                         .Where(s => s.Value.Skill?.Class >= PyjamaSpecialistClass)
                          .ToList())
             {
                 character.Skills.TryRemove(entry.Key, out _);
@@ -78,7 +78,7 @@ namespace NosCore.GameObject.Services.SkillService
         private async Task SendSkillListAsync(ICharacterEntity character, bool useSpecialist)
         {
             var ordered = character.Skills.Values
-                .Where(s => s.Skill != null && (s.Skill!.Class >= FirstSpecialistClass) == useSpecialist)
+                .Where(s => s.Skill != null && (s.Skill!.Class >= PyjamaSpecialistClass) == useSpecialist)
                 .OrderBy(s => s.Skill!.CastId)
                 .Select(s => s.SkillVNum)
                 .ToList();
--- a/src/NosCore.GameObject/Services/SkillService/SkillService.cs
+++ b/src/NosCore.GameObject/Services/SkillService/SkillService.cs
@@ -33,22 +33,102 @@ namespace NosCore.GameObject.Services.SkillService
                     (key, oldValue) => characterSkill);
             }
 
-            // Push the refreshed list to the client. Matches OpenNos GenerateSki:
-            // primary/secondary are the class starter vnums (200 + 20*Class, +1),
-            // followed by every learned skill ordered by cast id so the bar draws
-            // deterministically. Without this packet the client's hotbar stays empty
-            // and the server's skill-cast gate is invisible.
-            var classByte = (byte)character.Class;
+            await SendSkillListAsync(character, useSpecialist: false).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Past this class the specialist kits begin. Up to 31 are the playable jobs and a few
+        /// shared containers; from 32 upwards there is one class per transformation.
+        /// </summary>
+        private const int FirstSpecialistClass = 31;
+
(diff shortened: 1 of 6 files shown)


src/NosCore.GameObject/Services/SkillService/SkillService.cs
```diff
@@ -33,22 +33,102 @@ public async Task LoadSkill(ICharacterEntity character)
...
+        /// </summary>
+        private const int FirstSpecialistClass = 31;
+
+        public async Task LoadSpecialistSkillsAsync(ICharacterEntity character, short morph, byte spLevel)
+        {
+            // The previous card's skills go first: changing specialist without clearing would
+            // leave the kit of every card ever worn.
+            RemoveSpecialistSkills(character);
+
+            // WHICH CARD, THE FILE SAYS. Skill.dat, DATA section, second field: it carries the
```
erwan-joly: too much comments

denislauri1999: Done in e0a5953f: 79 lines of my comments removed, plus the 6 ConfigureAwait(false) for the same reason you gave on #2321. One Italian comment I had left in a test is now in English.

Build clean, 999 tests green.

src/NosCore.GameObject/Services/SkillService/SkillService.cs
```diff
@@ -33,22 +33,68 @@ public async Task LoadSkill(ICharacterEntity character)
...
-            // Push the refreshed list to the client. Matches OpenNos GenerateSki:
-            // primary/secondary are the class starter vnums (200 + 20*Class, +1),
-            // followed by every learned skill ordered by cast id so the bar draws
-            // deterministically. Without this packet the client's hotbar stays empty
-            // and the server's skill-cast gate is invisible.
-            var classByte = (byte)character.Class;
+            await SendSkillListAsync(character, useSpecialist: false);
+        }
+
+        private const int FirstSpecialistClass = 31;
```
erwan-joly: Is there a way to avoid hardcoding this ?

erwan-joly: also if there isn't we can have a better name with the name of the first specialist class instead

denislauri1999: Renamed in 689b00a8, and the number was wrong too.

Measured in Skill.dat on the Class column:

    class 31   721 skills, 386 with no English name at all
    class 32    11 skills, all named, first is "Pillow Fight" -> Pyjama
    class 33    11 skills, "Two-Handed Sword Attack"
    class 84    11 skills, "Frost Fist"

32..84 are the specialist cards, one per card; 31 is a bucket of monster and leftover skills. The specialist class is morph + 31, so the first card (Pyjama, morph 1) lands on 32.

It is now `PyjamaSpecialistClass = 32`. On avoiding it entirely: nothing in the files declares the boundary, so I would only be moving the number somewhere else — but this way the name is checkable against the data. Note `> 31` was also inconsistent with the two other comparisons, which used `>= 31` and so treated those 721 bucket skills as specialist ones; all three are `>= 32` now.

Build clean, 980 tests green.

src/NosCore.GameObject/Services/SkillService/SkillService.cs
```diff
@@ -33,22 +33,68 @@ public async Task LoadSkill(ICharacterEntity character)
...
+            {
+                character.Skills.TryRemove(entry.Key, out _);
+            }
+        }
+
+        private async Task SendSkillListAsync(ICharacterEntity character, bool useSpecialist)
+        {
             var ordered = character.Skills.Values
-                .Where(s => s.Skill != null)
+                .Where(s => s.Skill != null && (s.Skill!.Class >= PyjamaSpecialistClass) == useSpecialist)
```
erwan-joly, line 81: so none of the non specialist skill are after specialist one ?

denislauri1999: Right, that partition is the whole assumption, so I measured it rather than argue it. Counting every record in `Skill.dat` by its Class column:

```
class  0   193 skills   Rest, Pick Up, Open Shop        the adventurer, and the universal actions
class  1    55          Upgrade, ...                    swordsman
class  2    53          Capture, Archery, Dagger Stab   archer
class  3    49          Capture, Energy Bolt, Magic Shot  mage
class  4    50          Standard Punch, Fists of Fury   martial artist
class 27    33          Spiky Leather, Cower            class 28    75          Poisoned Arrow, Vine Arrow       |  partner/pet families,
class 29    30          Flaming Armour, Spectral Plague  |  not cards and not player classes
class 30    72          Cake Strike, Spinning Spout     /
class 31   721          386 of them with no English name at all   -> a leftover bucket
class 32    11          Pillow Fight, Play Dead, Sleep  ...                                                      >  53 classes, 627 skills, one CARD each
class 84    11          Frost Fist, Frost Rift          /
```

So: **nothing above 31 is anything other than a specialist card.** 32..84 is 627 skills across 53 classes, ~11 apiece — the outliers are 39 (3), 40 (9), 47 (6), 57 (20) and 70 (8), still one card each.

And nothing a character owns can land there: what comes out of `CharacterSkill` is classes 0..4, the specialist ones are put in by `LoadSpecialistSkillsAsync` from the card's morph. The specialist class is morph + 31, which is why the first card — Pyjama, morph 1 — is 32.

The one thing the boundary does exclude besides player skills is 27..30, and those are correctly out: they are partner and pet skill families, not cards.

PR #2327, opened 2026-08-27
A melee-only hit rate bonus fed the ranged rate too

denislauri1999:
## What

BCard type 4 declares three scopes and only the first was folded in — into **both** rates at once:

| subtype | client text |
|---|---|
| 11 / 12 | "Hit rate of **all** attacks is increased by %s." |
| 21 / 22 | "Hit rate of **melee** attacks is increased by %s." |
| 31 / 32 | "Hit rate of **ranged** attacks is increased by %s." |

So a melee-only bonus did nothing, and neither did a ranged-only one: **13 declarations** of 4/21 and 4/31 across items and cards had no effect at all.

Type 2 sitting next to it already keeps the three defences apart the same way, so this is the offensive half of a split that was half done, written to match.

## What is left out, and why

**4/41-42**, concentration during a magic attack. `CombatStats` has no field for it, and exposing one that nothing reads would be worse than the gap.

## Testing

- Builds with **0 warnings**; full suite green — **996 tests**, five of them new.
- The new tests assert the negative half as well as the positive: a melee bonus must leave `DistanceRate` **unchanged**, and a ranged bonus must leave `HitRate` unchanged. A leak into the other field raises nothing, so that is the assertion that earns its place.
- **Not verified in a client.** What somebody with a client would need to check - kept here rather than in `manual-test-plan.md`, since it belongs to this one change:

  - [ ] A buff with subtype 21 makes melee attacks land more often and leaves ranged alone
  - [ ] A buff with subtype 31 makes ranged attacks land more often and leaves melee alone
  - [ ] A subtype 11 buff raises both

## Summary by CodeRabbit

* **Bug Fixes**
  * Melee and ranged hit-rate bonuses are now applied only to their respective attack ranges.
  * Bonuses affecting all attack ranges continue to apply to both melee and ranged calculations.
  * Increasing and decreasing hit-rate effects now stack and subtract correctly.

* **Tests**
  * Added coverage for melee, ranged, all-range, decreasing, and stacked hit-rate effects.

### DIFF
--- a/documentation/manual-test-plan.md
+++ b/documentation/manual-test-plan.md
@@ -69,6 +69,12 @@ Boss skills declaring 90% should take most of a bar in one hit.
 - [ ] With both in play nothing locks up or throws — the attacker's guarantee is resolved
       first by design
 
+### Hit rate by scope — BCard type 4
+
+- [ ] A buff with subtype 21 makes melee attacks land more often and leaves ranged alone
+- [ ] A buff with subtype 31 makes ranged attacks land more often and leaves melee alone
+- [ ] A subtype 11 buff raises both
+
 ### Elemental resistance — BCard types 13 and 14
 
 - [ ] A buff that raises your resistance to an element reduces the elemental damage you
--- a/documentation/manual-test-plan.md
+++ b/documentation/manual-test-plan.md
@@ -69,12 +69,6 @@ Boss skills declaring 90% should take most of a bar in one hit.
 - [ ] With both in play nothing locks up or throws — the attacker's guarantee is resolved
       first by design
 
-### Hit rate by scope — BCard type 4
-
-- [ ] A buff with subtype 21 makes melee attacks land more often and leaves ranged alone
-- [ ] A buff with subtype 31 makes ranged attacks land more often and leaves melee alone
-- [ ] A subtype 11 buff raises both
-
 ### Elemental resistance — BCard types 13 and 14
 
 - [ ] A buff that raises your resistance to an element reduces the elemental damage you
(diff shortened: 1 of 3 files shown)


documentation/manual-test-plan.md
```diff
@@ -69,6 +69,12 @@ Boss skills declaring 90% should take most of a bar in one hit.
 - [ ] With both in play nothing locks up or throws — the attacker's guarantee is resolved
       first by design
 
+### Hit rate by scope — BCard type 4
+
+- [ ] A buff with subtype 21 makes melee attacks land more often and leaves ranged alone
```
erwan-joly: this file should be removed those are for conversation in git pr

erwan-joly: actually there is useful stuff there so maybe we should just keep the thing that are relevant for 1 pr in the conversation and keep the global useful one like mounts, combat etc

erwan-joly:
need to rebase

documentation/manual-test-plan.md
denislauri1999: Done — the three checks moved into the PR description, and the file keeps only what is useful across PRs.

On the second half: master's copy still carries six sections that belong to one PR each (types 40, 37/31, 16, 13/14, 34/35, and the combo/equipment ones are the general kind you mean). Happy to prune those in a separate small PR and leave Setup, Combat basics, Mounts, Class change, Equipment gates and the regression sweep — say the word and I'll open it.

erwan-joly (dismissed):

erwan-joly:
need rebase

documentation/manual-test-plan.md
denislauri1999: Settled from my side — master's copy is now the six cross-PR sections only (Setup, Combat, Monsters, Character, Regression sweep, Not verifiable yet), which is the split you described.

One thing still pulls both ways, and it will cost another round trip if it stays: `CLAUDE.md` says **"Add to that file when you add a gameplay feature"**, while this thread says a PR's own checks belong in the PR conversation. For a new feature those two point in opposite directions — the checks that prove it are exactly the per-PR ones.

My reading is that the file should hold what stays true across features and the PR should hold what proves this one, so a new feature adds to the file only when it creates a *lasting* check. If that is what you mean, the `CLAUDE.md` line could say so; if you mean something else, I would rather have it from you than guess.

PR #2329, opened 2026-08-27
Name every declared effect the client describes

denislauri1999:
## What

`BCard.dat.md` listed **231** declared effects with no `BCardEffect` member. **207** now have one, taken from the client sentence the row already carries.

The **24** left are the ones the client declares with **no sentence at all**, and that is not just an English gap: their tokens resolve in **none of the eight languages** the client ships. A name for those would be an invention rather than a reading, so they stay unnamed. `EveryDeclaredEffectIsNamed` still lists them, and it is opt-in for exactly that reason.

## Three the sentences forced

Each of these would raise nothing if read the other way round, which is why they are worth naming out loud.

**Type 32, subtypes 21 and 31 both say MP recovery goes up.** 21 is in points, 31 is a percentage:

```
21: MP recovery is increased by %s.
31: MP recovery is increased by %s%%.
```

so the second is `RecoveryMpRecoveryPercentIncreased` — the same distinction the file already makes between types 13 and 94.

**Type 60, subtypes 21 and 51 carry identical text** ("The Rage bar increases by %s%%.") *and* identical unresolved labels, so nothing in the file separates them. Named by position rather than by a meaning the file does not give.

**Type 77 subtype 42 is not the mirror of 41.** Both regenerate the caster's max HP, and 42 adds "This effect does not affect the caster" — so it is `SummonAndRecoverHpRegenerateFromCasterMaxHpNotCaster`, not a `Negated`.

## Worth a separate look

`Type89Unknow = 8921`, but that row's client text is "Provides a %s%% probability to reset the cooldown of %s." Left alone here to keep this PR to additions.

## Testing

- Builds with **0 warnings**; full suite green.
- `NosCore.Parser.Tests` passes **104/104** with `TestCategory!=OPTIONAL-TEST`, which is how CI runs it. Worth knowing: `dotnet test NosCore.sln` **silently skips that project** on my machine — I only saw the vocabulary tests by invoking the project directly.
- Every name was read against the sentence in its own row. No behaviour changes: nothing reads these yet, which is the point.

## Summary by CodeRabbit

- **Documentation**
  - Expanded effect descriptions across combat, capture, raids, specialists, elements, fishing, tokens, and skills.
  - Clarified and corrected several effect definitions, including positioning, item interactions, damage effects, armour, gravity, fuel, and guaranteed outcomes.
  - Added clearer wording for negated effects, missing-health effects, and elemental resistance effects.

- **Improvements**
  - Introduced consistent names for previously unnamed effects.
  - Clarified elemental increase/decrease and resistance percentage effects for easier interpretation.


documentation/dat/BCard.dat.md
erwan-joly, line 9: i don't really think we need all those value here the explanation at the top should suffice

PR #2332, opened 2026-08-28
Multiplied attack and defence, which nothing folded in

denislauri1999:
BCard types 34 and 35 — "attack power is multiplied by %s" and "defence is multiplied by %s" — are declared in `BCardEffect` and were absent from the fold in `BattleStatsProvider`. A card promising a fivefold defence changed nothing, and nothing raised.

They are the only two families that state a **factor** rather than a percentage. Their values are small whole numbers, so reading one as a percentage would turn the strongest card of the family into five per cent; and the decreasing half says "divided by", so it halves rather than subtracting two.

"All" and the per-kind subtype add into one factor, the way the flat halves of types 3 and 9 already do.

**Left out, deliberately**

* `35/51-52` — "damage is reduced by %s%% per debuff stack, up to %s%%": a rate and a ceiling, not a factor, and there is no count of stacks to feed it.
* `34/41-42` — `CombatStats` has no magic attack field for them to land in.
* `34/51-52` — these scale with missing HP, which is a different shape again.

**What was tested**

* 10 unit tests in `MultiplyingStatTests`: the "all" subtype reaching all three defences, each per-kind subtype reaching only its own, the attack factor moving both ends of the roll, the halving, the identity at 1, and "all" plus a kind adding into one factor.
* Measured red-before by making the fold a no-op: **8 of the 10 failed**; with the fold in place, `NosCore.GameObject.Tests` is **452 passed, 0 failed**, and `dotnet build NosCore.sln` is 0 warnings, 0 errors.
* **No in-game verification** — I did not run a client for this. Checks added to the combat section of `documentation/manual-test-plan.md`; they are deliberately coarse, because a fivefold defence should be visible in a single exchange rather than needing measurement.

## Summary by CodeRabbit

* **New Features**
  * Added support for multiplying attack and defence stats through factor-based effects.
  * Supports overall and melee, ranged, or magic-specific defence adjustments.
  * Negative factors reduce stats by division, while a factor of one leaves them unchanged.

* **Documentation**
  * Added a manual testing checklist for multiplied attack and defence effects.

### DIFF
--- a/documentation/manual-test-plan.md
+++ b/documentation/manual-test-plan.md
@@ -77,6 +77,19 @@ Boss skills declaring 90% should take most of a bar in one hit.
 - [ ] An "all elements" effect applies to each of fire, water, light and dark rather than
       being ignored
 
+### Multiplied attack and defence — BCard types 34 and 35
+
+These state a factor, not a percentage, so the change is meant to be obvious rather than
+subtle: a card worth 5 is fivefold.
+
+- [ ] A buff carrying "defence is multiplied" makes the same monster's blows land for a
+      small fraction of what they did a moment earlier
+- [ ] The melee, ranged and magic subtypes each move only their own defence: check with
+      three attackers, or one attacker switching weapons
+- [ ] A buff carrying "attack power is multiplied" makes your own numbers jump by whole
+      multiples, not by a few per cent
+- [ ] The decreasing half halves rather than taking a couple of points off
+
 ---
 
 ## Monsters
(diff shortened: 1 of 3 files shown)


documentation/manual-test-plan.md
```diff
@@ -77,6 +77,19 @@ Boss skills declaring 90% should take most of a bar in one hit.
 - [ ] An "all elements" effect applies to each of fire, water, light and dark rather than
       being ignored
 
+### Multiplied attack and defence — BCard types 34 and 35
```
erwan-joly, line 80: i missed this file likely this should be in conversation on git not in a md file

PR #2358, opened 2026-08-30
fix: primary keys are discovered from one Npgsql model, not throwaway InMemory contexts

erwan-joly:
Architecture-review PR 5: the DI/startup cleanup.

## Problem

Three startup paths (PersistenceModule.DiscoverDaoMappings, MasterServerBootstrap's DAO registration, ParserBootstrap's) discovered each entity's primary key by constructing a **fresh InMemory `NosCoreContext` per property per DTO type** inside a `Where` lambda — hundreds of full EF model builds before the first request, against an InMemory model that lacks relational conventions, and the sole reason `Microsoft.EntityFrameworkCore.InMemory` shipped in the WorldServer, MasterServer, Parser and Database production assemblies.

## Change

- `PersistenceModule.FindPrimaryKeyProperty(Type)`: one lazily built **Npgsql** model (model building never opens a connection — the placeholder connection string is never dialed) answers every PK lookup; it's the same model the runtime uses
- Master and Parser bootstraps call the shared helper; their DTO filters are deliberately untouched
- `EFCore.InMemory` removed from all four `src` csprojs — it remains in test projects, where it belongs

This is the first slice of the larger DI consolidation (single container, DiGenerator for Master/Login) — kept separate because it's independently verifiable and removes the test package from prod images immediately.

## Verification

Build clean; Database.Tests, Parser.Tests, GameObject.Tests pass. Startup smoke of Master/Parser paths is covered by the parser's migration-and-import run whenever you next use it — the DAO registration path is identical in kind, only the PK source changed.

## Summary by CodeRabbit

- **Performance**
  - Improved application startup by streamlining database metadata processing.
  - Reduced resource usage during server and parser initialization.

- **Reliability**
  - Standardized database object registration across application components.
  - Improved consistency when identifying database records and their primary keys.

- **Maintenance**
  - Simplified database configuration by removing unnecessary in-memory database support.

### DIFF
--- a/src/NosCore.Database/Hosting/PersistenceModule.cs
+++ b/src/NosCore.Database/Hosting/PersistenceModule.cs
@@ -101,11 +101,31 @@ public sealed class PersistenceModule : Autofac.Module
         services.AddSingleton<IDao<IItemInstanceDto?, Guid>, Dao<ItemInstance, IItemInstanceDto?, Guid>>();
     }
 
+    // One real Npgsql model shared by every primary-key lookup at startup. Model
+    // building never opens a connection, and the InMemory provider this replaces
+    // produced a model without relational conventions while dragging a test-only
+    // package into every server.
+    private static readonly Lazy<Microsoft.EntityFrameworkCore.Metadata.IModel> DesignModel = new(() =>
+    {
+        var options = new DbContextOptionsBuilder<NosCoreContext>()
+            .UseNpgsql("Host=design-time-only", o => o.UseNodaTime()).Options;
+        using var context = new NosCoreContext(options);
+        return context.Model;
+    });
+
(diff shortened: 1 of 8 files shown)


erwan-joly:
Applied — but by deleting the call sites rather than null-checking them.

Both bootstraps were re-implementing the entity match and PK lookup that `PersistenceModule.DiscoverDaoMappings` already performs, and that method already skips entities whose key it cannot resolve. They now iterate it directly, so the null-forgiving operators are gone rather than guarded.

That also fixed a filter drift the duplication was hiding: MasterServer selected DTOs by name, excluding anything containing "InstanceDto" unless it also contained "Inventory". `ScriptedInstance : IStaticEntity` is not an item instance, so `ScriptedInstanceDto` was silently unregistered there. The shared filter excludes `IItemInstanceDto` by type instead.

PR #2365, opened 2026-08-30
fix(npc): the monster info card carries the monster's name

denislauri1999:
`GenerateNpcInfo` never assigns `Name`, so the field serialises as `-` and the
client draws a dash where the monster's name belongs. It compares field 25
against `@` and falls back when it does not match.

`Unknown` already defaults to `-1` in the packet class, so the rest of the line
was correct: `... 1360 630 -1 -`. The name is the only missing piece.

The `Replace(' ', '^')` is deliberate and meant to be deleted. `Name` is the
last field of the packet, and a last field is only escaped when it declares
`EscapeSpaces`. `EInfoNpcMonsterPacket.Name` does not, so the escaping has to
happen before the value is handed to the serialiser. Adding the flag belongs in
NosCore.Packets; once it ships, this line goes away. It matters in practice:
929 of the 1109 monster-info lines in a real capture carry a name with a space
in it, across 630 distinct names.

### What was tested

- A unit test on the **serialised** line, not on the object: it asserts the tail
  reads `-1 <name>` and that a spaced name arrives escaped. Written red first -
  without the assignment it fails on the dash.
- `dotnet build`: 0 warnings. `dotnet test NosCore.sln`: green, except
  `BCardVocabularyTests.EveryDeclaredEffectIsNamed`, which is marked
  `[TestCategory("OPTIONAL-TEST")]`, fails on `master` too, and is untouched by
  this branch.

### What was NOT tested

Not played. The NosCore servers are not started here, so there is no screenshot
of the info card. The client-side behaviour above is read from the client
binary, not observed.

## Summary by CodeRabbit

* **Bug Fixes**
  * NPC information packets now correctly preserve localized names, including names containing spaces.
  * English and French NPC names are serialized with the appropriate localization.

* **Tests**
  * Added coverage for NPC packet formatting, field counts, portrait data, and localized names.

### DIFF
--- a/src/NosCore.GameObject/Ecs/Extensions/NpcInfoExtensions.cs
+++ b/src/NosCore.GameObject/Ecs/Extensions/NpcInfoExtensions.cs
@@ -1,4 +1,4 @@
-//  __  _  __    __   ___ __  ___ ___
+﻿//  __  _  __    __   ___ __  ___ ___
 // |  \| |/__\ /' _/ / _//__\| _ \ __|
 // | | ' | \/ |`._`.| \_| \/ | v / _|
 // |_|\__|\__/ |___/ \__/\__/|_|_\___|
@@ -41,9 +41,7 @@ public static class NpcInfoExtensions
             DarkResistance = npc.DarkResistance,
             MaxHp = npc.MaxHp,
             MaxMp = npc.MaxMp,
-            // The serializer escapes a string field against the separator that follows it, and
-            // nothing follows this one, so the spaces have to go before it is handed over.
-            Name = npc.Name[language].Replace(' ', '^'),
+            Name = npc.Name[language],
         };
     }
 
--- a/src/NosCore.GameObject/Ecs/Extensions/NpcInfoExtensions.cs
+++ b/src/NosCore.GameObject/Ecs/Extensions/NpcInfoExtensions.cs
@@ -13,15 +13,6 @@ namespace NosCore.GameObject.Ecs.Extensions;
 
 public static class NpcInfoExtensions
 {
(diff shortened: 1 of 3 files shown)


src/NosCore.GameObject/Ecs/Extensions/NpcInfoExtensions.cs
```diff
@@ -50,6 +41,9 @@ public static EInfoNpcMonsterPacket GenerateNpcInfo(this NpcMonsterDto npc, Regi
             DarkResistance = npc.DarkResistance,
             MaxHp = npc.MaxHp,
             MaxMp = npc.MaxMp,
+            // The serializer escapes a string field against the separator that follows it, and
+            // nothing follows this one, so the spaces have to go before it is handed over.
+            Name = npc.Name[language].Replace(' ', '^'),
```
erwan-joly: don't need that replace we have a way to do it in the packets (serializer)

erwan-joly: Done in d270df9a — the flag existed, I just had to use it.

EscapeSpaces is on PacketIndexAttribute and the serializer already honours it for a last field; GInfoPacket, MlInfoBrPacket and MlintroPacket all declare it. EInfoNpcMonsterPacket.Name did not, which is what the Replace was standing in for. Added there in NosCoreIO/NosCore.Packets#509, released as 21.1.2, and this branch now pins it and hands the name over as it is.

The line tests are unchanged and still assert Mother^Cuby on the wire — they now prove the serializer does the escaping. Full suite green on 21.1.2 (1128 tests).

Also wrote the rule into CLAUDE.md in #2367 so the next one goes to the package instead of the call site.

